PR #466, opened 2018-03-02
CUMULUS-159 Fix PDR progress report on the dashboard

jennyhliu:


yjpa7145:
Are there tests to be updated or created for this?

jennyhliu:
There are only a couple of tests which don't apply to this change.  I'm not sure how to create tests for the dashboard.

scisco:
we don't have any real tests for the dashboard yet. @laurenfrederick could this be something for the next PR?

jennyhliu:
Added tests.  Please review and approve.

test/utils/pdrs.js
```diff
@@ -0,0 +1,48 @@
...
+
+test('test pdr-progress.js getProgress', function (t) {
+  const result = getProgress(d);
+  t.equal(result.percentCompleted, 25);
+  t.equal(result.percentFailed, 75);
+  t.equal(result.granulesCompleted, '4/4');
+  t.end();
+});
+
+test('test pdrs.js tableRow', function (t) {
```
yjpa7145: Could this test be simplified to:

```javascript
t.equal(col[4](d), 4);
t.equal(col[5](d), d.stats.processing);
t.equal(col[6](d), d.stats.failed);
t.equal(col[7](d), d.stats.completed);
t.end();
```

jennyhliu: yes. less code, less error.  Why didn't I think this way

PR #471, opened 2018-03-30
CUMULUS-394 Adding expanding errors

kkelly51:


package.json
```diff
@@ -84,6 +84,7 @@
     "node-bourbon": "^4.2.8",
     "node-notifier": "^4.5.0",
     "node-sass": "^3.7.0",
+    "react-collapsible": "^2.0.4",
```
sethvincent: should be in dependencies rather than devDependencies

kkelly51: Or both? Why did CI pass with it in devDependencies?

PR #474, opened 2018-05-31
Cumulus-561 

kkelly51:
Added correct log level format

All logs coming from elastic search have {level: 30} instead of {level: 'info'}

### DIFF
--- a/app/scripts/components/logs/viewer.js
+++ b/app/scripts/components/logs/viewer.js
@@ -145,10 +145,10 @@ var LogViewer = React.createClass({
             if (d.level) {
               if (d.level === 30) level = 'info';
               else level = d.level.toString().toLowerCase();
-            }
-            else {
+            } else {
               level = 'info';
             }
+
             return <p
               key={d.key}
               className='logs__item'><span className='logs__item--date'>{d.displayTime}</span> <span className={'logs__item--level logs__item--' + level}>{level}</span> {text}</p>;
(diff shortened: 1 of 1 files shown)


app/scripts/components/logs/viewer.js
```diff
@@ -141,7 +141,11 @@ var LogViewer = React.createClass({
             if (text.length > 200) {
               text = text.slice(0, 200) + '...';
             }
-            const level = d.level ? d.level.toLowerCase() : 'info';
+            let level;
+            if (d.level) {
```
yjpa7145: What happens if `d.level` is not set?

app/scripts/components/logs/viewer.js
```diff
@@ -141,7 +141,11 @@ var LogViewer = React.createClass({
...
+              if (d.level === 30) level = 'info';
+              else level = d.level.toString().toLowerCase();
```
scisco: I think this still fails if the level is integer but has a value other than 30.

app/scripts/components/logs/viewer.js
kkelly51: Good point, added that case.

app/scripts/components/logs/viewer.js
kkelly51: It won't fail, It will just output the number as a string instead of "error" or whatever else. 
`40 2018-05-21T19:27:31.417Z	ed63b375-e577-4a65-8142-dd8c25b5b6b8	{"level":"info","timestamp":"2018-05-21T19:27:31.407Z","sender":"undefined","message":"Received the provider { provider: 's3_provider' }"}`

PR #485, opened 2018-08-21
[CUMULUS-690] As a DAAC operator, I want to view the task name and version of all code run for a workflow

abarciauskas-bgse:
Summary:
* Adds text describing how to view task name, version and ARN
* Add test for task name, version and ARN text on `execution-status` page

Test plan:
- [x] Tests pass


app/scripts/components/executions/execution-status.js
```diff
@@ -135,6 +135,7 @@ var ExecutionStatus = React.createClass({
       <section className='page__section'>
         <div className='heading__wrapper--border'>
           <h2 className='heading--medium with-description'>Events</h2>
+          <p><b>More details</b> includes details of the step function event execution, including the tasks' name and version. To find all task name and versions, select <b>More details</b> for the last Lambda- or Activity-type event. There you should find a key / value pair "workflow_tasks" which lists all tasks' version, name and arn.</p>
```
mennovandiermen: Since this depends on Cumulus 1.9.1+ and the appropriate message adapters, maybe add a note about 'If task/version tracking is enabled (...)'. Also, it is worth remembering that `workflow_tasks` is only updated as each task executes, so only the last output will have information about all the tasks that have run.

kkelly51: Good point about the version,
And you do say the last lambda should have them all so that's fine, it just a little difficult to find. What if we pulled it out somewhere?

kkelly51:
Should we have a changelog entry for this? Since not a lot has changed on the front end but you did add testing?

app/scripts/components/executions/execution-status.js
abarciauskas-bgse: @mennovandiermen good point, added a note here: https://github.com/nasa/cumulus-dashboard/pull/485/commits/0f0c09253bd90315d8dac37929ebd0e5a314b572

abarciauskas-bgse:
@kkelly51 👍 added change log entry https://github.com/nasa/cumulus-dashboard/pull/485/commits/54c802fca9e598b3d1efc7a258c774575a6b2265

app/scripts/components/executions/execution-status.js
abarciauskas-bgse: Maybe I'm misunderstanding but I did try and describe that the _final_ event should be inspected:

> To find all task name and versions, select <b>More details</b> for the last Lambda- or Activity-type event.

I removed the sentence that came before this sentence to make the instruction more obvious - let me know if you think this content will be sufficient in helping users find the information they need.

abarciauskas-bgse:
@mennovandiermen re https://github.com/nasa/cumulus-dashboard/pull/485#pullrequestreview-149015379 I don't think I understand:

> Storing them in `cumulus_meta` instead is an option but would be part of recreating the initial problem of overly big messages.

It seems like the size increase from adding `workflow_tasks` to `cumulus_meta` would be relatively minor unless the workflow had _many_ tasks. I did some napkin math let me know if this makes sense to you:
* [character limit for input step functions is 32,768 characters](https://docs.aws.amazon.com/step-functions/latest/dg/limits.html#service-limits-general)
* base cumulus remote message input to a step function is 129 characters
* the addition of "workflow_tasks": [] is 20 characters
* The number of characters in 1 workflow task entry is 162 characters
* (32,768 - 129 - 20)/162 = 201, so we could have 201 workflow tasks before reaching the step function input limit.

Sidebar: the "store as remote event" threshold is [10000 bytes](https://github.com/nasa/cumulus-message-adapter/blob/cd78130f8d4fd5a9118ff6635f7148868abdb7ca/message_adapter/message_adapter.py#L397-L398), I can't recall where this limit came from.

Do you think we could store them in `cumulus_meta` so we don't have the issue where the `workflow_tasks` don't appear in event outputs for events which exceeded the 10000 bytes threshold??

mennovandiermen:
I guess we can. I'm just hesitant to say "throw it in cumulus_meta" because of the big message concerns. I'm not sure how much more we have coming down the pipeline that should be visible on the dashboard.

Anyway, changes will need to be made to [message.js](https://github.com/nasa/cumulus/blob/master/packages/deployment/lib/message.js#L138) and the [message adapter](https://github.com/nasa/cumulus-message-adapter/blob/master/message_adapter/message_adapter.py#L116) to put workflow tasks in cumulus_meta and check for them there as well.

kkelly51:
So, I'm still not seeing all the task names and versions. 
For example for a ingest granule workflow, I only see the workflow_tasks in meta up to the LambdaFunctionScheduled for moveGranule. Which means I'm not seeing the version for moveGranule. 
Everything after that is too large and would need to be retrieved from s3. 

What is we just made a call for the last step "ExecutionSucceeded" so we could make sure to get all the tasks? 
What happened with the conversation with Menno? Are we moving that information to cumulus_meta? Do I need to update something to get the latest changes? 

Also just a comment, but seeing the version say "LATEST" isn't that helpful. Is there any way to get the actual version?

abarciauskas-bgse:
All of the comments that @kkelly51 has made are valid and have already been discussed. To make those changes we should postpone the front-end story (this one, CUMULUS-690) until the API returns more helpful data - that is task name and version data in `cumulus_meta` so it's present even in remove event message as well as more specific task versions. I think the latter is dependent on @Jkovarik work https://github.com/nasa/cumulus/pull/527

cc @mennovandiermen @laurenfrederick

PR #487, opened 2018-08-28
CUMULUS-691 view the Cumulus logs for a workflow in the dashboard

kkelly51:
Added page to execution with execution logs


mennovandiermen:
In https://github.com/nasa/cumulus/blob/master/packages/api/lib/response.js#L271 you will see that if there is any error in the call to `func(cb)` it will return the `Invalid Authorization Token` response.

PR #489, opened 2018-09-17
default selection is null when opening dropdown

nemreid:
The `Execute` dropdown modal populated its list with choices, but the workflow that is initially displayed is not set as the selected workflow in the back end. If the initial workflow is used, the command will fail as the selection would still be `undefined`, resulting in the following error:
```
Error: "Bad Request: Workflow doesn't exist: s3://system-bucket/deployment/workflows/undefined.json for undefined."
```
This change updates the modal to show an empty selection so that this behavior is represented in the UI, and an operator is forced to manually select an option. Switching to an explicit null selection also reduces the potential for accidental workflow starts (i.e. mistakenly clicking 'Confirm' instead of 'Cancel'). 
null_selection

### DIFF
--- a/app/scripts/utils/table-config/granules.js
+++ b/app/scripts/utils/table-config/granules.js
@@ -84,7 +84,7 @@ export const simpleDropdownOption = function (config) {
       options={config.options}
       id={config.label}
       onChange={config.handler}
-      noNull={true}
+      noNull={false}
     />
   );
 };


markdboyd:
Talked to @mennovandiermen. My only concern here is if the modal goes away when you click "Confirm" without making a selection but the API doesn't actually succeed, then you aren't notified of that. I think a potential solution would be to make the workflow selection required

nemreid:
This PR is very stale. Closing and re-submitted the issue as CUMULUS-1743.

PR #497, opened 2018-10-19
CUMULUS-649

markdboyd:
- Add test for attempting to access Rules page when not logged in
- Add test for viewing rules when logged in
- Add login command to easily simulate login across multiple tests
  - Cypress recommends having one E2E test for login that performs the UI interaction for login, but [for subsequent tests to login programatically](https://docs.cypress.io/guides/references/best-practices.html#Organizing-Tests-Logging-In-Controlling-State)

PR #498, opened 2018-10-22
CUMULUS-913|914: Add tests to configure collection and provider

jennyhliu:
Added tests to configure collection and provider


cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,82 @@
...
+      cy.get('@collections').should('have.attr', 'href', '#/collections');
+      cy.get('@collections').click();
+
+      cy.url().should('include', 'collections');
+      cy.get('.heading--xlarge').should('have.text', 'Collections');
+
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
+    });
+
+    it('collections page displays a button to add a new collection', () => {
```
flamingbear: So now I think the AC aren't that good. Below I asked if the tests fulfilled the AC. The same thing occurs here.
`filling in the new collection form and clicking submit displays the list of collections with the new collection in it`


But clicking submit doesn't take you to the list of collections, it takes you to the collection detail page (/collections/collection/MOD09GQ/006) and you would have to follow to `Back to Collections` to see the added collection.

But is this even possible with the FAKE-API we're running?

Is there a way we can mock a new collection and add it and if we could, is that useful in any way?

flamingbear: This also doesn't appear to pass the command line tests in circle (headless).  I tried it from my command line and saw it fail but also see that it works running cypress interactively.

jennyhliu: It's weired.  The command line worked for me.  But I added one more check to see if it helps.

jennyhliu: @laurenfrederick @kkelly51 The fake api server uses static data, and doesn't do anything with the POST/PUT/DELETE request.  Is it enough that we only test the interface?  To satisfy the AC, we need to use real api server.  Or we could update the fake api server to act like a real server, and I'm not sure how much work is there and if it's worth to do that.

markdboyd: I'm not sure if you want to do this, but using Cypress you can actually stub routes on our fake API or in general. See https://docs.cypress.io/guides/guides/network-requests.html#Requests.

I'm doing this in [#499](https://github.com/nasa/cumulus-dashboard/pull/499/files#diff-18ed6a1b0092f55d3cb4dd831481484aR68).

cypress/support/commands.js
```diff
@@ -23,3 +23,12 @@
 //
 // -- This is will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add('login', (host) => {
```
markdboyd: I think it will be helpful if we can get #499 merged and use the `login` command added there. Cypress recommends having one login flow test using the actual UI, but then for subsequent tests to make whatever network requests/app state changes are necessary to have someone logged in so subsequent tests can be much faster

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,91 @@
...
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+      cy.get('.heading--large').should('have.text', 's3_provider');
+      cy.url().should('include', '#/providers/provider/s3_provider');
+    });
+
+    it('provider page has buttons to edit or delete the provider', () => {
+      cy.visit(`${host}/#/providers/provider/s3_provider`);
+      cy.get('.heading--large').should('have.text', 's3_provider');
+      cy.get('a').contains('Edit').should('exist').as('editprovider');
+      cy.get('@editprovider').should('have.attr', 'href').and('include', '#/providers/edit/s3_provider');
```
markdboyd: When we have lots of chained calls like this, can we split them up onto indented new lines for readability?

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,91 @@
+describe('Dashboard Providers Page', () => {
+  const host = process.env.DASHBOARD_HOST || 'http://localhost:3000/';
+
+  describe('When not logged in', () => {
+    it('should redirect to login page', () => {
+      cy.visit(`${host}#/providers`);
```
markdboyd: If we can get #497 merged in, you won't have to preprend the host to each `visit` call. So this will just become `cy.visit('#/providers')`

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+      cy.get('.heading--large').should('have.text', 'Collection Overview');
+      cy.get('a').contains('Add a Collection').should('exist').as('addCollection');
+      cy.get('@addCollection').should('have.attr', 'href', '#/collections/add');
+      cy.get('@addCollection').click();
+
+      // fill the form and submit
+      // we need to use fake collection under test/fake-api-fixtures/collections, so the
+      // verification would work
+      const collection = '{{}"name":"MOD09GQ","version":"006","dataType":"MOD09GQ"}';
```
markdboyd: Is this intentional: `{{}`?

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,83 @@
+describe('Dashboard Collections Page', () => {
+  const host = process.env.DASHBOARD_HOST || 'http://localhost:3000/';
+
+  describe('When not logged in', () => {
+    it('should redirect to login page', () => {
+      cy.visit(`${host}#/collections`);
+      cy.url().should('include', '/#/auth');
```
markdboyd: I create a [re-usable assertion](https://github.com/nasa/cumulus-dashboard/pull/497/files#diff-bbe16e111202074abf127c887faa6d32R1) for this in #497 to make it easier to re-test this

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,83 @@
...
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
+
+      // delete collection
+      cy.get('button').contains('Delete').should('exist').click();
+      cy.get('button').contains('Confirm').click();
+
+      // goes back to collections list
+      cy.url().should('include', 'collections');
+      cy.get('.heading--xlarge').should('have.text', 'Collections');
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
```
markdboyd: The length of the collections list was greater than 1 before this collection was deleted, correct? So I don't think asserting that the table length is still greater than 1 really proves that this specific collection was removed from the list. Perhaps we could do a `should('not.exist')` on an element in the table we expect to exist for this collection?

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,83 @@
...
+      cy.get('a').contains('Edit').should('exist').as('editCollection');
+      cy.get('@editCollection').should('have.attr', 'href').and('include', '#/collections/edit/MOD09GQ/006');
+      cy.get('@editCollection').click();
+
+      cy.get('.heading--large').should('have.text', 'Edit MOD09GQ___006');
+
+      // cy.clear doesn't work for textarea, maybe it's related to https://github.com/cypress-io/cypress/issues/2056
+      // const collection = '{{}"name":"MOD09GQ","version":"006","dataType":"MOD09GQ"}';
+      // cy.get('textarea').clear({force: true}).type(collection, {force: true});
+      cy.get('textarea').type(' ', {force: true});
```
markdboyd: What is the `type` command actually changing about the collection that we can test to verify it was updated?

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,83 @@
...
+      // we need to use fake collection under test/fake-api-fixtures/collections, so the
+      // verification would work
+      const collection = '{{}"name":"MOD09GQ","version":"006","dataType":"MOD09GQ"}';
+      cy.get('textarea').type(collection, {force: true});
+      cy.get('form').get('input').contains('Submit').click();
+
+      // displays the new collection
+      cy.get('.heading--xlarge').should('have.text', 'Collections');
+      cy.get('.heading--large').should('have.text', 'MOD09GQ / 006');
+      cy.url().should('include', '#/collections/collection/MOD09GQ/006');
```
markdboyd: Does this not work as `cy.url().should('equal', '#/collections/collection/MOD09GQ/006')`?

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,91 @@
...
+      cy.get('@providerinput').contains('Provider Name').siblings('input').type('s3_provider');
+      cy.get('@providerinput').contains('Concurrent Connnection Limit').siblings('input').clear().type(5);
+      cy.get('@providerinput').contains('Protocol').siblings().children('select').select('s3').should('have.value', 's3');
+      cy.get('@providerinput').contains('Host').siblings('input').type('cumulus-test-sandbox-internal');
+
+      cy.get('form div input[value=Submit]').click();
+
+      // displays the new provider
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+      cy.get('.heading--large').should('have.text', 's3_provider');
```
markdboyd: This assertion is failing for me locally

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+      cy.get('form div ul').as('providerinput');
+      cy.get('@providerinput').contains('Concurrent Connnection Limit').siblings('input').clear().type(12);
+      cy.get('@providerinput').contains('Host').siblings('input').clear().type('cumulus-test-sandbox-internal-new');
+
+      cy.get('form div input[value=Submit]').click();
+
+      // displays the updated provider
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+      cy.get('.heading--large').should('have.text', 's3_provider');
```
markdboyd: This assertion is failing for me

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+      // delete provider
+      cy.get('.dropdown__options__btn').click();
+      cy.get('span').contains('Delete').click();
+      cy.get('button').contains('Confirm').click();
+
+      // goes back to providers list
+      cy.url().should('include', 'providers');
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
```
markdboyd: Does this actually prove that the provider was deleted?

cypress/integration/collections_spec.js
jennyhliu: yes.  I will make updates after you merge.

cypress/integration/collections_spec.js
laurenfrederick: @markdboyd would this be instead of using the fake api? 

@jennyhliu  @flamingbear Is the problem that it doesn't take you back to the collection page or that the collection page does not have the new collection in it (or both)?

cypress/integration/collections_spec.js
jennyhliu: yes, to escape `{` https://docs.cypress.io/api/commands/type.html#Arguments

cypress/integration/collections_spec.js
jennyhliu: the `url()` starts with host

cypress/integration/collections_spec.js
jennyhliu: The clear() doesn't work, so the text isn't really changed.  I haven't found a way to update the text  without clear and retype

cypress/integration/collections_spec.js
jennyhliu: With current fake-api server, the data isn't changed after `delete` request

cypress/integration/providers_spec.js
jennyhliu: will update.  thanks

cypress/integration/collections_spec.js
markdboyd: @laurenfrederick In theory, yes, we could use that entirely instead of the fake API. However, I was just using it to stub certain routes that aren't going to provided as static fixtures (e.g. DELETE). 

Ultimately, I think it may be better to go with @jennyhliu's approach of just adding DELETE support to the fake API as long as we're going to use that

jennyhliu: @jennyhliu After the new collection `MOD09GQ/006` is added, it displays the collection page for the new collection.  I can add step to `Back to Collection` page if it's necessary.

cypress/integration/collections_spec.js
markdboyd: OK. I think that will change once #497 is merged

cypress/integration/collections_spec.js
markdboyd: So if we aren't changing the collection definition, how can test to ensure that the collection was updated? Even if we could clear/retype, would that help if we didn't change the collection definition?

cypress/integration/collections_spec.js
markdboyd: Ah, I see your point. One thing: if you delete the collection using the button from the list page instead of the detail page, the row does get removed and you can test against that. I'm doing this in #499.

If you want to stick with this approach though, we could get around the problem with Cypress stubbing for `cy.route`. This probably deserves a bigger conversation with @kkelly51 about how long we intend to keep the fake API and how we want to deal with faking dynamic data updates

cypress/integration/collections_spec.js
laurenfrederick: What I am thinking is we need to test that it's going to the "right" page, whatever page that is. I don't think the test that the collection page shows the new collection, since that's an API call and we're handling those with API unit and integration tests.

It just seems like the AC is off from what is actually happening on the dashboard?

fake-api.js
```diff
@@ -49,6 +130,20 @@ app.get('/auth', (req, res) => {
   res.status(200).end();
 });
 
+app.use('/', express.static('test/fake-api-fixtures', { index: 'index.json' }));
+
+app.post('/*', (req, res) => {
```
markdboyd: I don't think we need these if we're going to have specific routes for each data type

fake-api.js
```diff
@@ -31,15 +35,92 @@ function fakeApiMiddleWare (req, res, next) {
...
 }
 
-app.use('/', fakeApiMiddleWare );
-app.use('/', express.static('test/fake-api-fixtures', { index: 'index.json' }));
+const jsonParser = bodyParser.json();
+
+app.use('/', fakeApiMiddleWare);
+
+app.get('/collections', (req, res) => {
+  res.send(collectionsJson);
```
markdboyd: Does this work across test runs? Meaning if your test to delete a collection comes is the last test in your Cypress run and then you re-run the Cypress collection tests, I would think this list will still have that collection missing from the list.

To resolve this you'll have to update to the fake API storage method in #500

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,111 @@
...
+      cy.get('.heading--large').should('have.text', name);
+      cy.url().should('include', `#/providers/provider/${name}`);
+
+      // verify the new provider is added to the providers list
+      cy.get('a').contains('Back to Provider').click();
+      cy.get('.heading--large').should('have.text', 'Provider Overview');
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
+
+      // providers table doesn't have tr[data-value="${name}"] like collections page
+      cy.contains('table tbody tr a', name)
```
markdboyd: I think this is actually a better way to test rather than relying on the `data-value` attribute anyway

cypress/integration/providers_spec.js
markdboyd: How do we know for sure that the providers list should still be greater than 1 at this point?

markdboyd: I see, it is because the fixture for providers has 2 items? In which case, we could test for equality to 2 exactly. But I'm wondering whether it's even important to test the list length as long as we can assert that the row for the specific provider we deleted does not exist

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,111 @@
...
+
+      // displays the new provider
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+      cy.get('.heading--large').should('have.text', name);
+      cy.url().should('include', `#/providers/provider/${name}`);
+
+      // verify the new provider is added to the providers list
+      cy.get('a').contains('Back to Provider').click();
+      cy.get('.heading--large').should('have.text', 'Provider Overview');
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
```
markdboyd: If we wanted this assertion to provide a new provider was created, we would need to test for greater than 2, correct? Since the fixture for providers includes 2 items. But I feel like this assertion may be unnecessary anyway since line 65 really proves that the provider was added to the list

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,111 @@
...
+      cy.visit('/');
+
+      cy.get('nav li a').contains('Providers').should('exist').as('providers');
+      cy.get('@providers').should('have.attr', 'href', '#/providers');
+      cy.get('@providers').click();
+
+      cy.url().should('include', 'providers');
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+
+      cy.get('table tbody tr').its('length').should('be.gt', 1);
```
markdboyd: I feel like this should test for either > 1 OR exactly the length of our fixture

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,123 @@
...
+      cy.visit('/#/collections');
+      shouldBeRedirectedToLogin();
+    });
+  });
+
+  describe('When logged in', () => {
+    const name = 'TESTCOLLECTION';
+    const version = '006';
+    beforeEach(() => {
+      cy.login();
```
markdboyd: Ideally, I think we should also `cy.task('resetState')` here, as it means the tests are really well isolated and not dependent on each other's state

jennyhliu: I have used the `TESTCOLLECTION` through out the test suite.

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,123 @@
...
+      //
+      // ace editor: editor.setValue(collection) does update the value in the ace_content,
+      // but doesn't cause the textarea value to be updated, and the updated conent is not submitted
+      // with the form.
+      // const collection = '{"name":"TESTCOLLECTION","version":"006","dataType":"TESTCOLLECTION", "duplicateHandling":"replace"}';
+      // const escapedCollection = collection.replace('{', '{{}');
+      // cy.get('textarea')
+      //   .type('{selectall}{backspace}', { force: true })
+      //   .type(escapedCollection, { force: true });
+      // cy.get('.ace_editor').then((ele) => {
```
markdboyd: Even though this is commented out, I think we can remove it since I don't think this approach will ever work

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,123 @@
...
+      cy.visit(`/#/collections/collection/${name}/${version}`);
+      cy.contains('a', 'Edit').should('exist').as('editCollection');
+      cy.get('@editCollection')
+        .should('have.attr', 'href')
+        .and('include', `#/collections/edit/${name}/${version}`);
+      cy.get('@editCollection').click();
+
+      cy.get('.heading--large').should('have.text', `Edit ${name}___${version}`);
+
+      // There are issues updating the react form which uses ace editor.
```
markdboyd: I'm not sure we need to include these comments. People can just reference the actually working approach here

fake-api.js
```diff
@@ -72,6 +130,10 @@ app.get('/auth', (req, res) => {
 
 app.use('/', express.static('test/fake-api-fixtures', { index: 'index.json' }));
 
+app.put('/*', (req, res) => {
```
markdboyd: Is this route actually being used?

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,106 @@
...
+
+      // fill the form and submit
+      const collection = '{{}"name":"TESTCOLLECTION","version":"006","dataType":"TESTCOLLECTION", "duplicateHandling":"error"}';
+      cy.get('textarea').type(collection, {force: true});
+      cy.get('form').get('input').contains('Submit').click();
+
+      // displays the new collection
+      cy.get('.heading--xlarge').should('have.text', 'Collections');
+      cy.get('.heading--large').should('have.text', `${name} / ${version}`);
+      cy.url().should('include', `#/collections/collection/${name}/${version}`);
```
markdboyd: Do we want to add assertions for the correct `duplicateHandling` or any of the other specific properties?

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,106 @@
...
+      cy.visit(`/#/collections/collection/${name}/${version}`);
+
+      // delete collection
+      cy.contains('button', 'Delete').should('exist').click();
+      cy.contains('button', 'Confirm').click();
+
+      // verify the collection is now gone
+      cy.url().should('include', 'collections');
+      cy.get('.heading--xlarge').should('have.text', 'Collections');
+      cy.get(`table tr[data-value="${name}___${version}"]`).should('not.exist');
```
markdboyd: Can we update this to `cy.contains('table tbody tr a', name).should('not.exist');` for consistency with the other tests?

cypress/integration/providers_spec.js
```diff
@@ -0,0 +1,112 @@
...
+        .select('s3', {force: true}).should('have.value', 's3');
+
+      cy.get('@providerinput').contains('Host').siblings('input').type('test-host');
+
+      cy.get('form div input[value=Submit]').click();
+
+      // displays the new provider
+      cy.get('.heading--xlarge').should('have.text', 'Providers');
+      cy.get('.heading--large').should('have.text', name);
+      cy.url().should('include', `#/providers/provider/${name}`);
```
markdboyd: Can we add some assertions for the specific provider property values?

jennyhliu: added assertion for property values of collection and provider

cypress/integration/collections_spec.js
```diff
@@ -0,0 +1,116 @@
...
+      cy.get('form').get('input').contains('Submit').click();
+
+      // displays the new collection
+      cy.contains('.heading--xlarge', 'Collections');
+      cy.contains('.heading--large', `${name} / ${version}`);
+      cy.url().should('include', `#/collections/collection/${name}/${version}`);
+
+      // verify the collection's properties by looking at the Edit page
+      cy.contains('a', 'Edit').should('exist').click();
+      cy.get('form .ace_content')
```
markdboyd: How are you sure that this is testing the state after the collection was edited and not before? I know that there is a "click" command above but I don't think it guarantees the page is fully reloaded before the next command. 

It seems like it might be preferable to test the contents of the "view" page and not the "edit" page

jennyhliu: @markdboyd Which "view" page are you preferring to?  The overview page of the collection doesn't the information we are looking for.  There is a `Definition` page which is similar to the `Edit` page.

markdboyd: Nevermind, I was misunderstanding

PR #499, opened 2018-10-22
CUMULUS-657

markdboyd:
* Adds test to verify that a rule can be deleted from the rules list page and that after deletion the rule no longer appears on the page


markdboyd:
I didn't add any assertion/test for `When not logged in, attempting to delete a rule fails` because I'm not sure what that would mean in this context. We already have a test verifying that you can't view the rules listing page if you aren't logged in and since the Delete button appears on the listing page, it would seem unnecessary to add another test for this case.

cypress/integration/rules.spec.js
```diff
@@ -61,5 +61,27 @@ describe('Rules page', () => {
...
+      cy.get('.form--controls button')
+        .contains('Delete')
+        .should('exist')
+        .click();
+      cy.get('.modal')
+        .should('exist')
+        .get('button')
+        .contains('Confirm')
+        .click();
+      cy.get('table tbody tr').should('not.exist');
```
jennyhliu: This won't work if we have multiple rules.

test/fake-api-db/index.js
```diff
@@ -0,0 +1,49 @@
...
+        return data;
+      })
+      .then((data) => seed(this.filePath, data));
+  }
+
+  getItems () {
+    return fs.readJson(this.filePath);
+  }
+
+  deleteItem (name) {
```
jennyhliu: This only works for `rule`

markdboyd: Yeah, I thought that storage for other types of data could override. Would you prefer I move this to the `FakeRulesDb` class?

jennyhliu: yes.  delete collection for example has `name and version` as argument.

markdboyd: OK. I moved it to the `FakeRulesDb` class

PR #500, opened 2018-10-25
CUMULUS-657

markdboyd:


test/fake-api-db/index.js
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+const rulesJson = require('../fake-api-fixtures/rules/index.json');
```
mennovandiermen: Worth noting using require on json config/input files is dangerous because of `require` caching the loaded object. If `rulesJson` is modified here, as it has been in previous commits, that modified object will be loaded in the next invocation that `require`s the same file. This has bitten us in the kinesis integration tests. Better to read in the file or `cloneDeep` it after requiring.

PR #501, opened 2018-10-29
Cumulus-645 Test creating a rule

kkelly51:
Wrote a test for creating a rule


cypress/integration/rules.spec.js
```diff
@@ -67,6 +67,26 @@ describe('Rules page', () => {
...
+      cy.get('textarea').clear({force: true});
+      const newRule = '{{}"name":"newRule","workflow":"HelloWorldWorkflow","provider":"PODAAC_SWOT","collection":{{}"name":"MOD09GQ","version":"006"},"meta":{{}},"rule":{{}"type":"onetime","value":""},"state":"ENABLED"}';
+      cy.get('textarea').type(newRule, {force: true});
+      cy.get('form').get('input').contains('Submit').click();
+
+      cy.get('.heading--xlarge').should('have.text', 'Rules');
+      cy.contains('table tbody tr a', 'newRule')
+        .should('exist')
+        .and('have.attr', 'href', '#/rules/rule/newRule');
+      cy.contains('table tbody tr a', 'PODAAC_SWOT')
```
markdboyd: This assertion doesn't prove anything because the existing rule from the fixture also has a provider of PODAAC_SWAT. As Jenny suggested, maybe we can keep the assertion to verify the rule was added to the list but if we want to test its properties we should look at the individual rule page

cypress/integration/rules.spec.js
```diff
@@ -67,6 +67,26 @@ describe('Rules page', () => {
...
 
+    it('creating a rule should add it to the list', () => {
+      cy.visit('/#/rules');
+      cy.get('a').contains('Add a rule').should('exist').as('addRule');
+      cy.get('@addRule').should('have.attr', 'href', '#/rules/add');
+      cy.get('@addRule').click();
+
+      cy.get('textarea').clear({force: true});
+      const newRule = '{{}"name":"newRule","workflow":"HelloWorldWorkflow","provider":"PODAAC_SWOT","collection":{{}"name":"MOD09GQ","version":"006"},"meta":{{}},"rule":{{}"type":"onetime","value":""},"state":"ENABLED"}';
+      cy.get('textarea').type(newRule, {force: true});
```
markdboyd: You need to make sure to clear the existing contents of the textarea otherwise you can end up typing at the end of the existing JSON which gives an error when you submit. 

screen shot 2018-10-29 at 10 31 56 am


```suggestion
      cy.get('textarea')
        .type('{selectall}{backspace}`, { force: true })
        .type(newRule, {force: true});
```

kkelly51: Hmmm the clear on line 76 was suppose to take care of that. It works for me like 80% of the time.

markdboyd: Yeah, we're just running into the same general issue where we need a reliable way of updating the contents of these textareas

cypress/integration/rules.spec.js
```diff
@@ -67,6 +67,26 @@ describe('Rules page', () => {
...
+      cy.get('form').get('input').contains('Submit').click();
+
```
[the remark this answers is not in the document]
kkelly51: Added

cypress/integration/rules.spec.js
kkelly51: Added a second clear makes it work in 100% of my test runs. Might seem silly but it gets up past the textarea issue right now

app/scripts/reducers/rules.js
```diff
@@ -59,8 +59,9 @@ export default function reducer (state = initialState, action) {
   const { data, id } = action;
   switch (action.type) {
     case RULE:
+      const foundRule = data.results.find((result) => result.name === id);
```
markdboyd: Why was this change necessary?

kkelly51: For some reason it was returning results[0] which was the original rule, not the one I just created. so the data filling the newRules page was from the wrong rule.

This even after I updated the fake-api only return the specified rule. but data.results had all rules in there so taking the first one wasn't necessarily correct

PR #502, opened 2018-10-30
CUMULUS-983 - Create reliable way to update JSON textareas

markdboyd:
- Adds `editJsonTextarea` Cypress command
- Update babel dependencies to v7
  - Cypress was throwing strange babel errors on `JSON.stringify` in the commands file when I added the `editJsonTextarea` command. Updating the babel dependencies was the only solution I could find. It also required me to update Ava to v1.0.0, but I verified that everything is still working
- Add missing assertions throughout tests and cleanup


app/scripts/components/form/text-area.js
```diff
@@ -33,6 +33,12 @@ const TextAreaForm = React.createClass({
     this.props.onChange(this.props.id, value);
   },
 
+  setEditorRef: function (editorRef) {
+    if (window.Cypress) {
+      window.aceEditorRef = editorRef;
```
markdboyd: This is what makes it all work: we're exposing the editor from the actual editor component, which allows us to use it to `getValue` and `setValue` directly instead of using unreliable approaches with `cy.type`.

cypress/integration/collections_spec.js
```diff
@@ -71,36 +78,35 @@ describe('Dashboard Collections Page', () => {
...
       cy.get('@editCollection').click();
 
-      cy.contains('.heading--large', `Edit ${name}___${version}`);
-
-      // There are issues updating the react form which uses ace editor.
-      // Neither cy .clear nor ace editor works.
-      // As a workaround, we add additional parameter at the end of collection metadata
-      const meta = '"meta": "testmetadata"}';
-      cy.get('textarea').type(`{backspace},${meta}`, { force: true });
+      cy.contains('.heading--large', `Edit ${name}___${version}`).should('exist');
```
markdboyd: You'll see a lot of changes like this. The reason why is that `cy.contains('.heading--large', `Edit ${name}___${version}`);` doesn't assert anything - it's just a selection method. So if we actually want to verify that these elements exist, we need to add `should` assertions

jennyhliu: If the content is not there, the `cy.contains` will fail due to timeout though.

jennyhliu: https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Default-Assertions

markdboyd: From the documentation you linked:

> cy.contains() expects the element with content to eventually exist in the DOM.

So `cy.contains` has built-in retry behavior just like `cy.get` and therefore I wouldn't expect it to fail more or less often. However, I also just saw this in that documentation:

> All DOM based commands automatically wait for their elements to exist in the DOM.
> You never need to write .should('exist') after a DOM based command.

So I was wrong and you don't actually need the `should('exist')`. I'll remove it throughout this PR

cypress/integration/collections_spec.js
```diff
@@ -71,36 +78,35 @@ describe('Dashboard Collections Page', () => {
...
-      cy.contains('.heading--xlarge', 'Collections');
-      cy.contains('.heading--large', `${name} / ${version}`);
+      cy.get('.heading--xlarge').should('have.text', 'Collections');
+      cy.get('.heading--large').should('have.text', `${name} / ${version}`);
 
       // verify the collection is updated by looking at the Edit page
       cy.contains('a', 'Edit').should('exist').click();
       cy.contains('form .ace_content', meta).should('exist');
-      cy.contains('.heading--large', `Edit ${name}___${version}`);
+      cy.get('.heading--large').should('have.text', `Edit ${name}___${version}`);
```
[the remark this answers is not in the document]
markdboyd: That's interesting. I think that `cy.get` and `cy.contains` should have the same retry behavior, but I'm not sure that I even made this change intentionally so I'll revert it

cypress/integration/collections_spec.js
```diff
@@ -71,36 +78,35 @@ describe('Dashboard Collections Page', () => {
...
-      // There are issues updating the react form which uses ace editor.
-      // Neither cy .clear nor ace editor works.
-      // As a workaround, we add additional parameter at the end of collection metadata
-      const meta = '"meta": "testmetadata"}';
-      cy.get('textarea').type(`{backspace},${meta}`, { force: true });
+      cy.contains('.heading--large', `Edit ${name}___${version}`).should('exist');
 
+      // update collection and submit
+      const meta = 'metadata';
+      cy.editJsonTextarea({ data: { meta }, update: true });
```
jennyhliu: Can you also update an existing field, such as set `duplicateHandling` to a different value?

markdboyd: Yeah, if you pass in this `update: true` parameter, then whatever you specify for `data` will be merged with the existing object using `_.assign` from Lodash: https://lodash.com/docs/4.17.10#assign

PR #503, opened 2018-11-01
CUMULUS-647

markdboyd:
* Add fake API fixture for workflows
* Add test for viewing workflows list in the dashboard


cypress/integration/workflows_spec.js
```diff
@@ -0,0 +1,34 @@
...
+      cy.visit('/');
+
+      cy.contains('nav li a', 'Workflows').as('workflows');
+      cy.get('@workflows').should('have.attr', 'href', '#/workflows');
+      cy.get('@workflows').click();
+
+      cy.url().should('include', 'workflows');
+      cy.contains('.heading--xlarge', 'Workflows');
+
+      cy.get('table tbody tr').its('length').should('be.eq', 3);
```
kkelly51: Could we get a few more checks for what we should be seeing on the workflow table?

PR #504, opened 2018-11-05
CUMULUS-990

markdboyd:
- Improve error reporting output
- Add logic to fake API DELETE routes for providers/collections to check for associated rules
- Add test for deleting a collection with an associated rule
- Add test for deleting a provider with an associated rule

### DIFF
--- a/app/scripts/components/collections/overview.js
+++ b/app/scripts/components/collections/overview.js
@@ -119,7 +119,7 @@ const CollectionOverview = React.createClass({
           <Link className='button button--small form-group__element--right button--green' to={`/collections/edit/${collectionName}/${collectionVersion}`}>Edit</Link>
           {lastUpdated(get(record, 'data.timestamp'))}
           {overview}
-          {errors.length ? <ErrorReport report={errors} /> : null}
+          {errors.length ? <ErrorReport report={errors} truncate={true} /> : null}
         </section>
         <section className='page__section'>
           <div className='heading__wrapper--border'>
--- a/app/scripts/components/errors/report.js
+++ b/app/scripts/components/errors/report.js
@@ -43,14 +43,14 @@ var ErrorReport = React.createClass({
         message = JSON.stringify(report);
       } else {
         message = report.message;
-        stack = report.stack ? report.stack.split('\\n').map(s => <p key={s}>{s}</p>) : null;
+        stack = report.stack
+          ? report.stack.split(`\n`).map((s, index) => <p key={index}>{s}</p>)
+          : null;
       }
       return (
         <div>
-          <p>
-            <strong key={message}>{name}: </strong>
-            {/* <Collapsible trigger={this.truncate(message)} triggerWhenOpen={message}/> */}
-          </p>
+          <strong key={message}>{name}: </strong>
+          <Collapsible trigger={this.truncate(message)} triggerWhenOpen={message}/>
           <Collapsible trigger={this.truncate(report.stack)} triggerWhenOpen={stack}/>
         </div>
       );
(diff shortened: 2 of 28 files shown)


app/scripts/components/collections/overview.js
```diff
@@ -119,7 +119,7 @@ const CollectionOverview = React.createClass({
           <Link className='button button--small form-group__element--right button--green' to={`/collections/edit/${collectionName}/${collectionVersion}`}>Edit</Link>
           {lastUpdated(get(record, 'data.timestamp'))}
           {overview}
-          {errors.length ? <ErrorReport report={errors} /> : null}
+          {errors.length ? <ErrorReport report={errors} truncate={true} /> : null}
```
jennyhliu: why do we truncate the errors?

cypress/integration/collections_spec.js
```diff
@@ -117,5 +117,24 @@ describe('Dashboard Collections Page', () => {
...
+      const version = '006';
+
+      cy.visit(`/#/collections/collection/${name}/${version}`);
+
+      // delete collection
+      cy.contains('button', 'Delete').click();
+      cy.contains('button', 'Confirm').click();
+
+      // error should be displayed indicating that deletion failed
+      cy.get('.error__report');
```
jennyhliu: add assertion to check the error message

cypress/integration/providers_spec.js
```diff
@@ -156,5 +156,26 @@ describe('Dashboard Providers Page', () => {
...
+
+      // delete provider
+      cy.get('.dropdown__options__btn').click();
+      cy.contains('span', 'Delete').click();
+      cy.contains('button', 'Confirm').click();
+
+      // error should be displayed indicating that deletion failed
+      cy.contains('Provider Overview')
+        .parents('section')
+        .get('.error__report');
```
jennyhliu: add assertion to check the error message

app/scripts/components/collections/overview.js
markdboyd: In a lot of cases, especially for errors with stack traces, the error output is really long. So what `truncate` actually does is initially show a shorter version of the error message with a "show more" link. And when the "show more" link is clicked, it expands to show the whole error message. Also, I added logic so that it won't truncate/add the "show more" link unless the error message is long enough to need it

app/scripts/components/errors/report.js
```diff
@@ -28,23 +28,53 @@ var ErrorReport = React.createClass({
...
       if (!report.message) {
         message = JSON.stringify(report);
       } else {
         message = report.message;
-        stack = report.stack ? report.stack.split('\\n').map(s => <p key={s}>{s}</p>) : null;
+        stack = report.stack
+          ? report.stack.split(`\n`).map((s, index) => <p key={index}>{s}</p>)
+          : null;
       }
-      return <div><p><strong key={message}>{name}: </strong> <Collapsible trigger={this.truncate(message)} triggerWhenOpen={message}/></p><Collapsible trigger={this.truncate(stack)} triggerWhenOpen={stack}/></div>;
```
markdboyd: I tested this out on several pages (dashboard homepage, collections list, etc) and it seemed like having the error message and error stack was quite redundant, because the error message was always included in the stack, but I welcome any feedback here

cypress/integration/collections_spec.js
markdboyd: I actually intentionally did not do that because it seemed kinda brittle. As long as these tests run against the fake API, it should be safe, though in the future they might run against our actual API?

jennyhliu:
When I test the cumulus dashboard against a real API, it appears the provider/collection can be deleted even if they are associated with a rule.  Do we have a separate ticket to verify cumulus server side behavior?

markdboyd:
Yeah Marc is working on implementing on the server side behavior in CUMULUS-656. The PR is open here: https://github.com/nasa/cumulus/pull/683

PR #509, opened 2018-11-09
CUMULUS-651: Edit provider and collection

jennyhliu:
The tests for Editting provider and collection have been added under PR.
https://github.com/nasa/cumulus-dashboard/pull/498

This PR has minor updates.

### DIFF
--- a/cypress/integration/collections_spec.js
+++ b/cypress/integration/collections_spec.js
@@ -9,6 +9,7 @@ describe('Dashboard Collections Page', () => {
       const name = 'MOD09GQ';
       const version = '006';
       cy.visit(`/#/collections/collection/${name}/${version}`);
+      shouldBeRedirectedToLogin();
     });
   });
 
--- a/cypress/integration/collections_spec.js
+++ b/cypress/integration/collections_spec.js
@@ -5,6 +5,10 @@ describe('Dashboard Collections Page', () => {
     it('should redirect to login page', () => {
       cy.visit('/#/collections');
       shouldBeRedirectedToLogin();
+
+      const name = 'MOD09GQ';
+      const version = '006';
+      cy.visit(`/#/collections/collection/${name}/${version}`);
     });
   });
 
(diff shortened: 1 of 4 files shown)


cypress/integration/collections_spec.js
```diff
@@ -43,14 +48,9 @@ describe('Dashboard Collections Page', () => {
...
       // fill the form and submit
-      const duplicateHandling = 'error';
-      const collection = {
-        name: 'TESTCOLLECTION',
-        version: '006',
-        dataType: 'TESTCOLLECTION',
-        duplicateHandling
-      };
-      cy.editJsonTextarea({ data: collection });
+      cy.fixture('TESTCOLLECTION___006.json').then((collection) => {
```
markdboyd: If the only value we're updating is the duplicateHandling, what is the value of adding a fixture with a full collection object? I'm a little hesitant to use Cypress fixtures because it implies that we're using Cypress request stubbing, which we are not doing.

jennyhliu: It is for `add collection`.  Since now adding a full collection object is possible (thank you!), we would want to add a full collection object complying with the collection schema.
https://github.com/nasa/cumulus/blob/master/packages/api/models/schemas.js#L143

cypress/integration/collections_spec.js
```diff
@@ -5,6 +5,10 @@ describe('Dashboard Collections Page', () => {
     it('should redirect to login page', () => {
       cy.visit('/#/collections');
       shouldBeRedirectedToLogin();
+
+      const name = 'MOD09GQ';
+      const version = '006';
+      cy.visit(`/#/collections/collection/${name}/${version}`);
```
markdboyd: Did you mean to add `shouldBeRedirectedToLogin` after this line?

jennyhliu: Yes.  It's for the acceptance criteria.

PR #510, opened 2018-11-13
CUMULUS-659: Tests for viewing workflow executions and their statuses

jennyhliu:


cypress/integration/executions_spec.js
```diff
@@ -0,0 +1,52 @@
...
+        .next().contains('li', '1 Failed')
+        .next().contains('li', '1 Running');
+
+      // shows a list of executions with IDs and status
+      cy.get('table tbody tr').as('list');
+      cy.get('@list').its('length').should('be.eq', 5);
+      cy.get('@list').each(($el, index, $list) => {
+        cy.wrap($el).children().as('children');
+        cy.get('@children').its('length').should('be.eq', 6);
+        cy.get('@children').eq(0).children('a').should('have.attr', 'href');
```
markdboyd: Can we assert the value of the href attribute?

markdboyd: I just realized that you're not doing this because the values will be different for each row. But can we just read the source JSON for the fake API fixture so we can compare against that? Right now these assertions are a little generic

PR #512, opened 2018-11-19
Cumulus-660 Test viewing an Execution and all details

kkelly51:
Test viewing an Execution and all details

AC:
When logged into the dashboard:
viewing the workflow executions page shows a link to execution details for each execution
viewing the workflow executions page and clicking on a link to execution details, it navigates to the execution details page
viewing the execution details page provides timing information for each step of the execution
viewing the execution details page provides status information for each step of the execution
viewing the execution details page provides overall status for the execution
viewing the execution details page provides overall timing for the execution
viewing the execution details page provides AWS ARNs for the execution and state machine
When not logged into the dashboard, detailed executions are not visible


cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,42 @@ describe('Dashboard Executions Page', () => {
...
+      cy.get('table tbody tr td[class=table__main-asset]').within(() => {
+        cy.get(`a[title=${executionName}]`).click({force: true});
+      });
+
+      cy.contains('.heading--large', 'Execution');
+      cy.contains('.heading--medium', 'Visual workflow');
+
+      cy.get('.table--wrapper')
+        .within(() => {
+          cy.get('tr[data-value=31]').children('td').as('items');
```
jennyhliu: Can we test all the steps for this execution, not just step 31?

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,42 @@ describe('Dashboard Executions Page', () => {
...
+          cy.get('@items').eq(0).should('have.text', '31');
+          cy.get('@items').eq(1).should('have.text', 'ExecutionSucceeded');
+          cy.get('@items').eq(2).invoke('text').should('match', /\d{2}:\d{2}:\d{2} \d{2}\/\d{2}\/\d{2}$/);
```
jennyhliu: Can we test the timestamp matches the one from fixture, even if they are in different format?

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,42 @@ describe('Dashboard Executions Page', () => {
...
+          cy.get('@items').eq(3).contains('More Details').click();
```
jennyhliu: Verify the actual content of the step payload

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,42 @@ describe('Dashboard Executions Page', () => {
...
+
+      cy.get('.status--process')
+        .within(() => {
+          cy.contains('Execution Status:').next().should('have.text', 'Succeeded');
+          cy.contains('Execution Arn:').next().should('have.text', executionArn);
+          cy.contains('State Machine Arn:').next().should('have.text', stateMachine);
+          cy.contains('Started:').next().invoke('text').should('match', /\d{2}:\d{2}:\d{2} \d{2}\/\d{2}\/\d{2}$/);
+          cy.contains('Ended:').next().invoke('text').should('match', /\d{2}:\d{2}:\d{2} \d{2}\/\d{2}\/\d{2}$/);
+        });
+    });
```
jennyhliu: Is `View Execution Logs` tested?

cypress/integration/executions_spec.js
kkelly51: We could, if you want. I just thought that seemed like a lot. There are a lot of rows.

cypress/integration/executions_spec.js
kkelly51: Added a few tests. Couldn't do a lot because all of the payloads are so different.

cypress/integration/executions_spec.js
kkelly51: Okay using the real fixture times in StartDate and stopDate. Didn't change this one because it's now in a loop with different times

cypress/integration/executions_spec.js
kkelly51: Testing multiple rows now.

cypress/integration/executions_spec.js
kkelly51: Added clicking on logs and a few tests on the logs page

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,59 @@ describe('Dashboard Executions Page', () => {
...
+        cy.get(`a[title=${executionName}]`).click({force: true});
+      });
+
+      cy.contains('.heading--large', 'Execution');
+      cy.contains('.heading--medium', 'Visual workflow');
+
+      cy.get('.table--wrapper')
+        .within(() => {
+          let i;
+          for (i = 26; i < 32; i++) {
```
markdboyd: I'm still not quite sure I understand why we're loading a fixture with ~32 events if we're only going to test against 5 of them. Can we just reduce the fixture to 5ish events and test against all of them? 
That way we can avoid "magic" numbers in our testing

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,59 @@ describe('Dashboard Executions Page', () => {
...
+            cy.get('@items').eq(3).contains('Less Details').click();
+          }
+        });
+
+      cy.get('.status--process')
+        .within(() => {
+          cy.contains('Execution Status:').next().should('have.text', 'Succeeded');
+          cy.contains('Execution Arn:').next().should('have.text', executionArn);
+          cy.contains('State Machine Arn:').next().should('have.text', stateMachine);
+          cy.contains('Started:').next().should('have.text', '20:05:10 11/12/18');
```
markdboyd: When I run this locally, I get a failure of `expected '<dd>' to have text '20:05:10 11/12/18', but the text was '15:05:10 11/12/18'`

jennyhliu: I saw this too.  I think the baseline code doesn't convert the timestamp string to another format correctly.

cypress/integration/executions_spec.js
markdboyd: How hard would it be to get the contents of the "execution details"  and "execution logs" areas, parse the JSON string, and compare directly against the fixtures? Seems like it would be a very complete test with less steps, but I haven't thought about it deeply so let me know if this seems like a bad idea

cypress/integration/executions_spec.js
jennyhliu: On the other side, we want to test if dashboard can display complex workflow execution.  Probably only `HelloWorld` workflow execution has only 5 events.  It would be interesting to see how long it will take to verify all the events.

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,59 @@ describe('Dashboard Executions Page', () => {
...
+          cy.contains('Ended:').next().should('have.text', '20:05:31 11/12/18');
+          cy.contains('Logs:').next()
+            .within(() => {
+              cy.get('a').should('have.attr', 'href', `#/executions/execution/${executionName}/logs`).click();
+            });
+        });
+
+      cy.contains('.heading--large', `Logs for Execution ${executionName}`);
```
markdboyd: It feels like this should be a separate test that starts with a `cy.visit` for the URL to the logs page. If nothing else, it breaks up this test which is already quite large

cypress/integration/executions_spec.js
markdboyd: If possible, I think we need to reduce the size of the fixture being loaded by the page/Cypress. At least on my machine, loading in this much data makes the test runner quite sluggish and less useful.

kkelly51: I thought we didn't want to change the data on the page. And there is no execution that only has a few events. Are you sure that's what you want to have happen?

cypress/integration/executions_spec.js
kkelly51: I think it would just as difficult to test everything directly with the fixture, because when the page is build it separate the data into different page elements, so we would still have to check each element. 
And on the other hand if we did just check that all the info was present on the overall page then that's not a very detailed test because we still want to make sure everything is where it is suppose to be.

cypress/integration/executions_spec.js
markdboyd: I think this is happening because `moment` defaults to the local timezone, which for all of us is east coast time. This command returns `15:05:10 11/12/18` for me:

`moment('2018-11-12T20:05:10.401Z').format('kk:mm:ss MM/DD/YY');`

To ensure that the test compares against the correct value for its timezone, we should import the `fullDate` function in `utils/format` that generates the timestamp for the dashboard and use it for the comparison in the test

cypress/integration/executions_spec.js
```diff
@@ -69,5 +69,92 @@ describe('Dashboard Executions Page', () => {
...
+
+      cy.contains('.heading--large', 'Execution');
+      cy.contains('.heading--medium', 'Visual workflow');
+
+      cy.get('.status--process')
+      .within(() => {
+        cy.contains('Execution Status:').next().should('have.text', 'Succeeded');
+        cy.contains('Execution Arn:').next().should('have.text', executionArn);
+        cy.contains('State Machine Arn:').next().should('have.text', stateMachine);
+        cy.contains('Started:').next().invoke('text').should('match', /\d{2}:05:10 11\/12\/18/);
```
markdboyd: Instead of this regex, can we do import the `fullDate` function in `utils/format` that generates the timestamp for the dashboard and use it for the comparison in the test? That should ensure that the comparison value is using the same timezone as the dashboard code to generate the date string

cypress/integration/executions_spec.js
```diff
@@ -69,5 +70,98 @@ describe('Dashboard Executions Page', () => {
...
+          if (index !== 0) {
+            cy.get('@columns').eq(3).contains(previousIdMatch);
+          }
+          cy.get('@columns').eq(3).contains('Less Details').click();
+        });
+      });
+    });
+
+    it('should show logs for a single execution', () => {
+      cy.contains('nav li a', 'Executions').click();
```
markdboyd: Having these assertions at the beginning of this test means it is dependent on the result of the previous test putting you on the right page. That could change as we refactor, so the more isolated we can make them, the better.

I would suggest moving these assertions and the check for the `Logs` link into the previous test. And starting this test with a command to visit the logs page:

`cy.visit(`/#/executions/execution/${executionName}/logs`);`

cypress/integration/executions_spec.js
```diff
@@ -69,5 +70,98 @@ describe('Dashboard Executions Page', () => {
...
       });
     });
+
+    it('should show a single execution', () => {
+      cy.contains('nav li a', 'Executions').as('executions');
+      cy.get('@executions').should('have.attr', 'href', '#/executions');
+      cy.get('@executions').click();
+
+      cy.url().should('include', 'executions');
+      cy.contains('.heading--xlarge', 'Executions');
```
markdboyd: These first few assertions all depend on the result of the previous test. They also seem redundant with the previous test. I would remove them and start this test with a `cy.visit` to the page for a single execution

PR #513, opened 2018-11-19
CUMULUS-989

markdboyd:
* [Added Auth0 library for managing JWTs](https://github.com/auth0/node-jsonwebtoken)
* Added state for JWTs
* Added actions/reducers for token refresh
* Added middleware to detect token expiration and request refresh
  * User is logged out if token refresh fails

Corresponding PR for the API: https://github.com/nasa/cumulus/pull/717


cypress/integration/executions_spec.js
```diff
@@ -37,8 +37,7 @@ describe('Dashboard Executions Page', () => {
         .next().contains('li', '1 Running');
 
       // shows a list of executions with IDs and status
-      const executionStatusFile = './test/fake-api-fixtures/executions/index.json';
-      cy.readFile(executionStatusFile).as('executionStatus');
+      cy.getFakeApiFixture('executions').as('executionStatus');
```
markdboyd: Completely unnecessary change but I wanted to centralize how we load test data from fake API fixtures so we don't have to update the path in multiple places when necessary

PR #524, opened 2018-12-06
Cumulus 1054 - Dashboard API Version Check

ifestus:
[CUMULUS 1054: Dashboard API Version Check](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1054)

**Summary:** I've added a `componentWillMount` function to the `app` component that will call two actions, one to grab cumulus-api version, and one to validate that api version against a list of compatible versions.

**Work in progress:**
* I'm looking for suggestions on how to improve my implementation - especially regarding the two actions I've added.
* I'd like to find a way to _not_ poll the api-version as much as it currently does...


app/scripts/components/app/index.js
```diff
@@ -2,25 +2,42 @@ import React from 'react';
...
+
   render: function () {
     return (
       <div className='app'>
         { target !== 'cumulus' ? (
           <div className='app__target--container'>
             <h4 className='app__target'>{displayCase(target)} ({displayCase(environment)})</h4>
+            <h5 className='app__api_version'>
+              {`API Version: ${this.state.api_version}`}
+              { this.state.version_compat_error ? 'INCOMPATIBLE CUMULUS API _ PLEASE CHECK' : '' }
```
markdboyd: I'm not seeing where this state is actually getting set? I think what you may want to do instead is read the API version from the `props` after connecting the API version state to this component.

app/scripts/reducers/api-version.js
```diff
@@ -0,0 +1,29 @@
+import {
+  API_VERSION,
+  API_VERSION_INFLIGHT,
+  API_VERSION_ERROR,
+  SET_API_VERSION_COMPAT_ERROR
+} from '../actions';
+
+export const initialState = {
+  api_version: '',
```
markdboyd: Nitpick: I think we've been sticking with a camelCase convention for store properties

app/scripts/actions/index.js
```diff
@@ -253,6 +253,29 @@ export const REFRESH_TOKEN_INFLIGHT = 'REFRESH_TOKEN_INFLIGHT';
...
+
+export const validateApiVersion = (apiVersion, compatibleList, dispatch) => {
+  const isCompatible = compatibleList.indexOf(apiVersion) > -1;
+  if (!isCompatible) {
+    dispatch({
+      type: SET_API_VERSION_COMPAT_ERROR,
+      payload: {
+        isCompatible,
+        apiVersion,
+        compatibleList
```
markdboyd: It doesn't seem like we need the compatibility list in this payload?

ifestus: Good point - I wanted to include that information in the error/warning and I wasn't sure if I wanted to write that error message in this function or in the reducer.

app/scripts/reducers/api-version.js
```diff
@@ -0,0 +1,29 @@
...
+  version_compat_error: false
```
markdboyd: I think it's easier to follow if this property is named the same as the action property that will update it, so changing it to `isCompatible`

app/scripts/components/app/header.js
```diff
@@ -25,7 +26,14 @@ var Header = React.createClass({
...
     api: React.PropTypes.object,
-    minimal: React.PropTypes.bool
+    minimal: React.PropTypes.bool,
+    apiVersion: React.PropTypes.object
+  },
+
+  componentWillMount: () => {
+    const { dispatch } = this.props;
+    dispatch(getApiVersion);
+    dispatch(validateApiVersion);
```
ifestus: * only do this dispatch if we're authenticated -`if this.props.api.authenticated`
* validateApiVersion can be done inside of the reducer
* could also look at `actions/index` mmt listCollections as an example

app/scripts/components/app/header.js
```diff
@@ -22,12 +22,18 @@ const paths = [
...
     api: React.PropTypes.object,
+    apiVersion: React.PropTypes.object,
+    dispatch: React.PropTypes.func,
+    location: React.PropTypes.object,
     minimal: React.PropTypes.bool
   },
 
+  componentWillMount: function () {
+    const { dispatch, api } = this.props;
+    if (api.authenticated) dispatch(getApiVersion(compatibleApiVersions));
```
markdboyd: Is there a reason to pass the `compatibleApiVersions` to the action rather than loading the variable directly in the actions file? It would just require less wiring between the component and the action

app/scripts/reducers/api-version.js
```diff
@@ -0,0 +1,35 @@
...
+export default function reducer (state = initialState, action) {
+  state = {...state};
+  switch (action.type) {
+    case API_VERSION:
+      state.versionNumber = action.payload.versionNumber;
+      state.warning = '';
+      break;
+    case API_VERSION_ERROR:
+      state.apiVersion = action.payload.error.message;
+      state.warning = 'Failed to acquire Cumulus API Version';
```
markdboyd: Do we want to warn the user if this check failed or just console.log it? If we do want to show the user, do we want to say something about how the compatibility cannot be determined?

fake-api.js
```diff
@@ -37,7 +39,7 @@ function fakeApiMiddleWare (req, res, next) {
     // respond with 200
     res.sendStatus(200).end();
     return;
-  } else {
+  } else if (publicEndpoints.indexOf(req.path) < 0){
```
markdboyd: I don't quite understand the reason for this change. It seems you're disabling the authorization check for the request to the `/version` endpoint on the fake API, but I'm not sure why?

app/scripts/reducers/api-version.js
ifestus: Currently this warning is shown on the front page just under the "Cumulus API Version."

fake-api.js
ifestus: I'm disabling the auth check in the `fake-api` for `/version` because it's not required to be authenticated agains the `default` api. As such, I'm not setting an `Authorization` header in the request from the dashboard to the `/version` endpoint

app/scripts/components/app/header.js
ifestus: Good call. I've updated the code.

app/scripts/actions/index.js
```diff
@@ -294,6 +300,54 @@ export const interval = function (action, wait, immediate) {
...
+};
+
+export const checkApiVersion = () => {
+  return (dispatch, getState) => {
+    const { versionNumber } = getState().apiVersion;
+    if (compatibleApiVersion.indexOf(versionNumber) < 0) {
+      dispatch({
+        type: API_VERSION_INCOMPATIBLE,
+        payload: {
+          warning: `Dashboard version incompatible with api version (${versionNumber})`
```
markdboyd: Nitpick: put api in all caps (API)

cypress/integration/main_page_spec.js
```diff
@@ -1,25 +1,44 @@
...
     shouldBeRedirectedToLogin();
   });
 
-  it('Logging in successfully redirects to the Dashboard main page', () => {
-    cy.get('div[class=modal__internal]').within(() => {
-      cy.get('a').click();
+  describe('When logged in', () => {
+    beforeEach(() => {
+      cy.visit('/');
+      cy.get('div[class=modal__internal]').within(() => {
```
markdboyd: This is super nitpicky, but in theory, we only need to test the manual login process (actually clicking the login link) once, whereas any other tests depending on logged in state can use the `cy.login` helper. 

So I think I might prefer if we move the "Logging in succesfully..." test back outside of this describe and use this code to login manually in it. Then in this block we can use `cy.login()` in `beforeEach` and leave your compatible API version number test as is.

PR #526, opened 2018-12-14
CUMULUS-1073

markdboyd:
Ostensibly, the goal of this ticket was to fix flakiness in a Cypress test for granules. However, I'm fairly sure that bug was exposing an edge case in our token refresh and request authorization flow. 

Ultimately, I felt that the best way to address that edge case (and other possible race conditions) was to introduce a new request middleware for adding request authorization and making requests. This is kind of what the previous `wrapRequest` function was doing, but because of how it was implemented, it was not possible to create chained actions on it.

So this ended up being a decent sized refactor, but I think it's appropriate/necessary for a few reasons:

- It allows us to distinguish which Redux actions are actually for API requests, instead of just checking whether actions are functions. Using the test of whether actions are functions for identifying API request actions is perilous because redux-thunk actions may also be functions but not API requests, which resulted in a lot of unnecessary code to avoid conflating the two.
- By moving the addition of the request authorization header to run in middleware **after the refresh token middleware**, we can eliminate possible race conditions involving deferred requests and expired tokens. I believe that this issue was the cause the flakiness in the test that we had to disable.
- It brings us more in line with intended redux-thunk usage patterns where you can have "action creators" that dispatch other actions and use promises to dispatch actions after other actions. 

**Other changes**

- I replaced the callback-style functions from `request` for making requests and replaced them with `request-promise`, which allow us to chain actions easily on API request actions. This is evident in the changes below to `listCollections` and allows us to remove previous workarounds.
- Added unit tests for the helper functions used in constructing requests and handling errors
- Added tests for the new middleware

**What to look for**

I'd especially like feedback on whether the test coverage here is sufficient. The refactor changes every action that makes an API request. We have Cypress tests that will cover some of these actions, but not all. The unit tests for the middleware do specifically test for the various request methods (GET, PUT, POST, DELETE). Since all the API request actions are just variations with different parameters, these tests may be sufficient, but I'm not sure.


app/scripts/actions/helpers.js
```diff
@@ -1,151 +1,67 @@
...
-  } else {
-    throw new Error('Must include a url with request');
-  }
-  config.json = true;
+export const configureRequest = (params = {}) => {
+  let config = params;
 
-  if (body && typeof body === 'object') {
-    config.body = body;
+  if (!config.url && !config.path) {
```
ifestus: This may be a bit easier to read, but completely up to you. This just gets rid of the need for a second `if (config.path)`
```nodejs
   if (!config.url || !config.path) {
   ...
    if (typeof config.path !== 'string') {
```

app/scripts/actions/index.js
```diff
@@ -249,34 +246,38 @@ export const API_VERSION_ERROR = 'API_VERSION_ERROR';
...
-      {
-        url: url.resolve(root, 'refresh')
-      },
-      { token }
-    ), (error, data) => {
-      if (error) {
+export const CALL_API = 'CALL_API';
+
+export const refreshAccessToken = (token) => {
+  return (dispatch) => {
```
ifestus: I like this formulation.

test/actions/store.js
```diff
@@ -0,0 +1,75 @@
...
+    .get('/test-path')
+    .reply(200, stubbedResponse);
+
+  const requestAction = {
+    type: 'TEST',
+    method: 'GET',
+    url: 'http://localhost:5001/test-path'
+  };
+  const actionObj = {
+    [CALL_API]: requestAction
```
ifestus: No qualms or changes needed, just curious as to the reason for `[ ]` as opposed to just `CALL_API`.

test/actions/helpers.js
```diff
@@ -1,44 +1,215 @@
...
+test('configureRequest() should maintain the request body', (t) => {
+  const requestBody = {
+    test: 'test'
+  };
+  const requestParams = {
+    url: 'http://localhost/test',
+    body: requestBody
+  };
+  const expectedConfig = {
+    ...t.context.defaultConfig,
```
ifestus: Is this value transformed into an array object at some point? I didn't think the spread operator works with dictionaries...

test/actions/store.js
```diff
@@ -0,0 +1,75 @@
...
+    Authorization: 'Bearer fake-token',
+    'Content-Type': 'application/json'
+  };
+
+  store.clearActions();
+});
+
+test.serial('dispatches TYPE_INFLIGHT and TYPE actions for API request action', async (t) => {
+  const stubbedResponse = { test: 'test' };
+  nock('http://localhost:5001')
```
ifestus: Good call with nock. I'm a fan. :+1:

app/scripts/actions/helpers.js
markdboyd: `if (!config.url || !config.path) {` would throw an error if you didn't have one or the other though, but it is valid to include either `url` or `path`.

test/actions/helpers.js
markdboyd: Yup, the spread operator does work on object literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals

test/actions/store.js
markdboyd: I think it's just a new syntax from ES6:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names

I also notice that these tests do not return the same thing:

```
var TEST = 'foo';
var test2 = { TEST: 'bar' }
console.log(Object.keys(test2))
```


```
var TEST = 'foo';
var test2 = { [TEST]: 'bar' }
console.log(Object.keys(test2))
```

Only in the second format is the object key actually recognized as the value of the literal (`foo`)

PR #527, opened 2018-12-17
Use babel to translate es6 operators in libraries.

flamingbear:
I was getting errors with libraries that used the spread operator but only with electron 59 headless (not chrome interactive).  Google suggested this would help and it appears to do so.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,9 @@
           "presets": [
             "@babel/preset-env",
             "@babel/preset-react"
+          ],
+          "plugins": [
+            "@babel/plugin-transform-spread"
           ]
         }
       ],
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -105,6 +108,7 @@
     "watchify": "^3.7.0"
   },
   "dependencies": {
+    "@babel/plugin-transform-spread": "^7.2.2",
```
markdboyd: I think maybe this should be a dev dependency like the other babel dependencies? You can save as a dev dependency using `yarn add @babel/plugin-transform-spread --dev`

flamingbear: Agreed, I saw where that landed, but this doesn't actually fix my problem.  I forgot to relink when I was testing, so I wasn't testing what I thought.

PR #528, opened 2018-12-19
Cumulus-1088 Update React on Dashboard

kkelly51:
Updated React from 15.6.2 -> 16.6.3

This had 2 Large breaks. 
* React.PropTypes~ no longer exists -> Now using PropTypes from 'prop-types'
* React.createClass no longer exists -> Now using import createReactClass from 'create-react-class'

error  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead


app/scripts/components/app/add-raw.js
```diff
@@ -48,13 +49,13 @@ const AddRaw = React.createClass({
     }
   },
 
-  componentWillMount: function () {
+  UNSAFE_componentWillMount: function () {
```
markdboyd: If we're going to go through the pain of the upgrade, I think we might as well move away from the deprecated methods. 

It seems like for this case, it would be recommended to use `getInitialState`. See https://reactjs.org/docs/react-without-es6.html#setting-the-initial-state

For other cases where we are just dispatching an action when the component is loading/loaded, it seems like `componentDidMount` is recommended. I don't love that, because `componentDidMount` fires slightly later than `componentWillMount` in the lifecycle, but perhaps it won't make a huge difference in practice

app/scripts/components/404.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 import React from 'react';
 import { connect } from 'react-redux';
+import createReactClass from 'create-react-class';
 
-var NotFound = React.createClass({
+var NotFound = createReactClass({
```
markdboyd: I was going to suggest that if we're going to bite of all the work of upgrading anyway, we might as well move from this style of declaring classes to use the new ES6 syntax. But that will be even more changes that have to happen at every component. It would probably be nice to do at some point, but perhaps we don't have to do it now

app/scripts/components/app/add-raw.js
```diff
@@ -48,13 +49,13 @@ const AddRaw = React.createClass({
...
     if (this.props.defaultValue) {
       this.setState({ data: JSON.stringify(this.props.defaultValue, null, 2) });
     }
   },
 
-  componentWillReceiveProps: function ({ state }) {
+  UNSAFE_componentWillReceiveProps: function ({ state }) {
```
markdboyd: It seems like the recommendation here is to upgrade to `componentDidUpdate`: https://reactjs.org/docs/react-component.html#unsafe_componentwillmount. Though the parameters of the function are different and some of the code running inside in these methods would need to be re-factored

app/scripts/components/app/add-raw.js
markdboyd: On further review, the upgrade paths to the new lifecycle methods are not necessarily drop-in replacements, so perhaps for the sake of keeping this PR sane we should leave it as is and file a separate ticket for the re-factor to use the new lifecycle methods. Thoughts?

kkelly51: Yes, agree that we should definitely take care of this. They did say this was going away completely after React v.17.
But for the sake of this PR not getting larger let's make that another ticket.

markdboyd: Cool. Would you mind writing up that ticket?

PR #534, opened 2018-12-22
Cumulus-1049: Add test to execution detail page for execution not in SFN API

jennyhliu:


cypress/integration/executions_spec.js
```diff
@@ -158,5 +158,59 @@ describe('Dashboard Executions Page', () => {
...
+      cy.get('.status--process')
+      .within(() => {
+        cy.contains('Execution Status:').next().should('have.text', 'Succeeded');
+        cy.contains('Execution Arn:').next().should('have.text', executionArn);
+        cy.contains('State Machine Arn:').next().should('have.text', stateMachine);
+        cy.contains('Started:').next().should('have.text', startMatch);
+        cy.contains('Ended:').next().should('have.text', endMatch);
+      });
+      /*
+      cy.get('table tbody tr').as('events');
```
markdboyd: Can this commented code be removed?

cypress/integration/executions_spec.js
```diff
@@ -158,5 +158,59 @@ describe('Dashboard Executions Page', () => {
...
+      cy.get('@executions').should('have.attr', 'href', '#/executions');
+      cy.get('@executions').click();
+
+      cy.url().should('include', 'executions');
+      cy.contains('.heading--xlarge', 'Executions');
+      const executionName = 'b313e777-d28a-435b-a0dd-f1fad08116t1';
+      const executionArn = 'arn:aws:states:us-east-1:596205514787:execution:TestSourceIntegrationIngestAndPublishGranuleStateMachine-yCAhWOss5Xgo:b313e777-d28a-435b-a0dd-f1fad08116t1';
+      const stateMachine = 'arn:aws:states:us-east-1:596205514787:stateMachine:TestSourceIntegrationIngestAndPublishGranuleStateMachine-yCAhWOss5Xgo';
+      cy.get('table tbody tr td[class=table__main-asset]').within(() => {
+        cy.get(`a[title=${executionName}]`).click({force: true});
```
markdboyd: Is there any reason to not do `cy.visit` directly to the URL for the execution? I think we already have a test that does a click on a row in the execution list, so we don't need to re-test that

cypress/integration/executions_spec.js
```diff
@@ -158,5 +158,51 @@ describe('Dashboard Executions Page', () => {
...
+      cy.contains('.heading--large', 'Execution');
+      cy.contains('.heading--medium', 'Visual workflow').should('not.exist');
+
+      const startMatch = fullDate('2018-12-06T19:18:11.174Z');
+      const endMatch = fullDate('2018-12-06T19:18:41.145Z');
+
+      cy.getFakeApiFixture(`executions/status/${executionArn}`).as('executionStatus');
+
+      cy.get('.status--process')
+      .within(() => {
```
markdboyd: Can we indent this whole `within()` block for readability?

cypress/integration/executions_spec.js
```diff
@@ -158,5 +158,51 @@ describe('Dashboard Executions Page', () => {
...
+        cy.contains('State Machine Arn:').next().should('have.text', stateMachine);
+        cy.contains('Started:').next().should('have.text', startMatch);
+        cy.contains('Ended:').next().should('have.text', endMatch);
+
+        cy.get('@executionStatus').its('execution').then((execution) => {
+          cy.contains('Input:').next().find('pre').then(($content) =>
+            // parse and stringify JSON string to get the same format as in the fixture
+            expect(JSON.stringify(JSON.parse($content.text()))).to.eq(execution.input));
+
+          cy.contains('Input:').next().contains('Show Input').click();
```
markdboyd: What do these two lines test?

jennyhliu: just to test the 'collapsible' elements are there.  I noticed the second 'click' doesn't work

markdboyd: Is there any CSS class or something that we can check instead of clicking the elements?

jennyhliu: There is `.Collapsible` class, but `Hide Input` won't show up unless `Show Input` is clicked.  I removed the click on `Hide Input/Ouput`, I'm not sure why it doesn't work with cypress, but it's not necessary for this ticket.

PR #535, opened 2019-01-02
CUMULUS-1087

markdboyd:
Currently, login failures may result in a state where the dashboard login URL contains a `?token=failed-token` query parameter. This query parameter then gets included in the `state` query parameter of the login link so that when the API redirects back to the dashboard even after successful login, the query parameter will be something like `?token=valid-token?token=failed-token`, which will confuse the dashboard's token parsing logic and cause a login failure.

To fix this problem, I've updated the code so that a `?token` query parameter is never included in the login link URL. And I've added tests to ensure the scenario is fixed.


markdboyd:
No, I only tested against the fake API since we can't access our real API for now. However, all of the prototyping and testing of the token authentication happens with the fake API anyway

PR #541, opened 2019-01-09
Move Cumulus API version to the footer

laurenfrederick:
Move the API version string from the header to the footer.


markdboyd:
These changes look fine, but we should re-name the file and update the tests in `test/components/header/version.js` to act on the footer instead of the header component?

markdboyd:
Can we also re-name `test/components/header/version.js` to `test/components/footer/version.js`?

PR #543, opened 2019-01-18
CUMULUS-1075

markdboyd:
- Fix issues with `yarn run build` that began after update to gulp v4
- Split up code build steps as separate job and add new job to CircleCI to test `yarn run build` which only runs on `master` branch
  - My original plan was to make the test process run the dashboard using `yarn run build` instead of `yarn run serve`, but for some reason this made the tests a lot more flaky. I tried to debug the issues but it was difficult for several reasons including that sourcemaps are not generated for `yarn run build`. So for now I propose leaving the test running code as is.
  - Since the goal here was just to add something to CI that tests `yarn run build`, this solution satisfies that goal and preserves our test stability


markdboyd:
It looks like the CI tests are never being reported as completed because we need to change some settings in the repo: https://circleci.com/docs/2.0/workflows-waiting-status/. But others should give feedback on this approach generally before we go changing any repo settings

jennyhliu:
Any reason we don't want to test `yarn run build` for non-master branches?

PR #544, opened 2019-01-21
add audit to circleci

jennyhliu:
circleci log should show the audit command and result.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,17 +7,6 @@ jobs:
     steps:
       - checkout
 
-      - run:
-          name: Run audit-ci
-          command: |
-            export NVM_DIR="/opt/circleci/.nvm"
-            [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-            nvm install v8.11.4
-            nvm alias default v8.11.4
-            npm install -g npm
-            npm install --no-audit
-            ./node_modules/.bin/audit-ci --moderate
-
       # restore cache
       - restore_cache:
           keys:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -50,6 +50,13 @@ jobs:
                 sleep 5
             done
 
+      - run:
+          name: run-audit-ci
+          command: |
+            nvm install-latest-npm
+            npm install --no-audit
```
kkelly51: Does npm install usually do an audit? 
And if so what's the difference between that and audit-ci ?

jennyhliu: `npm install` does audit but doesn't return non-zero status if there is any audit errors.  `audit-ci` allows you specify a vulnerability level and returns non-zero status if there are vulnerabilities  found on that level or high.

.circleci/config.yml
```diff
@@ -7,6 +7,17 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: Run audit-ci
+          command: |
+            export NVM_DIR="/opt/circleci/.nvm"
+            [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+            nvm install v8.11.4
```
kkelly51: What version were we using before? And why do we need this one ?

jennyhliu: For the `machine` executor, the existing`nvm --version` is 0.31.0, `node --version` is v6.1.0, `npm --version` is 5.4.2, and we need to update the node to 8 and npm to 6 for the audit.

.circleci/config.yml
```diff
@@ -7,6 +7,17 @@ jobs:
...
+            nvm alias default v8.11.4
+            npm install -g npm
+            npm install --no-audit
```
markdboyd: Is this installing the dependencies? If so, it's kind of duplicating effort since a later step does a `yarn install`. But if we can only run the audit through npm, I'm not sure what we can do

jennyhliu: Ok.  Updated to use `yarn` for audit, and fixed one package reported by `yarn` audit.

jennyhliu: Image `node:8.11.4` has `yarn` version `1.6.0` and `>=1.12.3` for audit to work.  The `node:10.13.0` has `yarn 1.12.3`.

I got permission error if I update the yarn version in docker container.  `docker run --rm -v /home/circleci/.cache:/root/.cache -v $(pwd):/home --workdir /home node:8.11.4 /bin/bash -c  'npm install -g yarn && yarn audit-ci --cache-folder /root/.cache'`
`/bin/bash: /usr/local/bin/yarn: Permission denied`
So I still need to run audit-ci outside of the container.

PR #546, opened 2019-01-29
Cumulus-1097-1 

kkelly51:
Removing all UNSAFE_componentWillMount

### DIFF
--- a/app/scripts/components/app/add-raw.js
+++ b/app/scripts/components/app/add-raw.js
@@ -40,9 +40,9 @@ class AddRaw extends React.Component {
     }
   }
 
-  UNSAFE_componentWillMount () { // eslint-disable-line camelcase
+  componentDidMount () {
     if (this.props.defaultValue) {
-      this.setState({ data: JSON.stringify(this.props.defaultValue, null, 2) });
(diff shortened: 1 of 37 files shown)


app/scripts/components/app/add-raw.js
```diff
@@ -40,9 +40,9 @@ class AddRaw extends React.Component {
     }
   }
 
-  UNSAFE_componentWillMount () { // eslint-disable-line camelcase
+  componentDidMount () {
     if (this.props.defaultValue) {
-      this.setState({ data: JSON.stringify(this.props.defaultValue, null, 2) });
+      this.setState({ data: JSON.stringify(this.props.defaultValue, null, 2) }); // eslint-disable-line react/no-did-mount-set-state
```
markdboyd: So it seems like this pattern is actually discouraged because it can cause extra rendering: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-mount-set-state.md.

In this case (and for other similar cases), I'm not sure why would we couldn't set `data: JSON.stringify(this.props.defaultValue, null, 2` in the constructor?

PR #547, opened 2019-02-04
Cumulus-1097-2 Remove all unsafe methods

kkelly51:
Removed all UNSAFE_componentWillReceiveProps

### DIFF
--- a/app/scripts/components/app/edit-raw.js
+++ b/app/scripts/components/app/edit-raw.js
@@ -54,9 +54,9 @@ class EditRaw extends React.Component {
     this.props.dispatch(getSchema(this.props.schemaKey));
   }
 
-  UNSAFE_componentWillReceiveProps ({ pk, state, schema, schemaKey }) { // eslint-disable-line camelcase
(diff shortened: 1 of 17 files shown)


app/scripts/components/app/edit-raw.js
```diff
@@ -89,6 +91,35 @@ class EditRaw extends React.Component {
     }
   }
 
+  // static getDerivedStateFromProps (nextProps) {
```
markdboyd: Do we need this code block?

app/scripts/components/app/edit.js
```diff
@@ -66,6 +67,35 @@ class EditRecord extends React.Component {
     }
   }
 
+  // static getDerivedStateFromProps (nextProps, prevState) {
```
markdboyd: Seems like we can remove this

kkelly51: oh no, that was a previous attempt. I kept it in just in case you wanted to see that I had tried doing it the other way. But I can remove it.

PR #549, opened 2019-02-08
Update Reconciliation Component

ifestus:
Make quick reference changes to reconciliation-report.js component to facilitate changes made in the Cumulus API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ The Cumulus Dashboard can be built inside of a Docker container, without needing
 
 Example of building for the "production" environment:
 ```
-$ ./bin/build_in_docker production
+$ ./bin/build_in_docker.sh production
 ```
 
 The compiled files will be placed in the `dist` directory.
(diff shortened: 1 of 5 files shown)


ifestus:
I'm checking whether or not more work needs to go into this. The page that lists reconciliation reports is not affected by the api changes, but the page that allows you to dive into each reconciliation report (clicking on the filename of a report) requires changes.

One thing I need to check, yet, is the `metaAccessors` on line 21 in `scripts/components/reconciliation-reports/reconciliation-report.js` - `okFileCount` no longer exists in the `record.data` object, it exists in two places: `record.data.filesInCumulus.onlyInS3` and `record.data.filesInCumulus.onlyInDynamoDb` objects.

jennyhliu:
image
The left botton corner has `.2`, and almost all the pages have this, not sure why it's there.  The `master` branch doesn't have it.

jennyhliu:
> image
> The left botton corner has `.2`, and almost all the pages have this, not sure why it's there. The `master` branch doesn't have it.

ifestus:
> I realized that the text is `Cumulus API Version: 1.11.2`, and it can't be read from the pages except the home page.

Oh dang, that's an issue. I can create another ticket for that, and let me know if you disagree, but I think it's out of scope for this specific ticket.

markdboyd:
Replaced by #559

PR #550, opened 2019-02-13
Cumulus-1149 Update Dashboard to reflect File format changes 

kkelly51:


app/scripts/components/granules/granule.js
```diff
@@ -44,16 +44,13 @@ const tableHeader = [
...
-  const key = chunks.slice(3).join('/');
+const makeLink = (bucket, key) => {
   return `https://${bucket}.s3.amazonaws.com/${key}`;
 };
 
 const tableRow = [
-  (d) => d.name || '(No name)',
-  (d) => d.filename ? (<a href={makeLink(d.filename)}>{d.filename ? link : nullValue}</a>) : null,
+  (d) => d.fileName || '(No name)',
+  (d) => (d.bucket && d.key) ? (<a href={makeLink(d.bucket, d.key)}>{d.fileName ? link : nullValue}</a>) : null,
```
yjpa7145: There is a slight change here.  Instead of the link text being `s3://some-bucket/path/to/my-file.txt`, it is now going to say `my-file.txt`.  I don't have a problem with that change, but want to make sure you're aware of it.

kkelly51: I'm a little confused. The link text is just "Link". Would you like to make it the fileName? In that case it would just be the same as the filename column to the left.

yjpa7145: I guess I'm confused, what is this ternary operator checking for? `d.fileName ? link : nullValue`?  If there is a fileName then the link text should be ___, otherwise the link text should be ___.  What are the blanks there?

markdboyd: Looks like if there is a `fileName`, it uses "Link" as the text, otherwise `nullValue`

PR #555, opened 2019-03-05
Workflow to deploy to Sandbox and SIT

laurenfrederick:
CUMULUS-1105
When merging to the deploy branch, deploy the dashboard to a sandbox bucket and a SIT bucket.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,3 +127,113 @@ jobs:
               --workdir /home \
               cypress/base:8 \
               yarn cypress-ci
+
+  deploy:
+    machine: true
+    working_directory: ~/cumulus-dashboard
+    steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -127,3 +127,113 @@ jobs:
...
+  version: 2
+  build_test_deploy:
+    jobs:
+      - build
+      - deploy:
+          requires:
+            - build
+          filters:
+            branches:
+              only: deploy
```
markdboyd: When are we ever going to have branches named `deploy`?

laurenfrederick: oh shoot i am a dum dum. I totally meant develop! Will update now. Wow, nice catch, thank you!

PR #559, opened 2019-03-15
CUMULUS-820: Update reconciliation reports

markdboyd:
- Update reconciliation report test data fixture to match new structure from API (as of v1.11.3)
- Add information from Cumulus vs CMR reconciliation report to page (files, collections, granules)
- Re-factor reconciliation report output for readability
- Add Cypress tests for new reconciliation report output
- Add new `ReportTable` component (with component tests) to simplify code
- Fix issue with alignment of API version warning in the footer


markdboyd:
I also just ran this against my API for my test stack and it worked without issue. However, my files in S3 table had over 11000 rows, which does not seem ideal because you have to scroll for quite a long time to see the rest of the tables. 

However, would this happen for a real deployment? Or is it only happening on my test stack because we have lots of extraneous files in S3?

app/scripts/components/reconciliation-reports/reconciliation-report.js
```diff
@@ -75,35 +68,119 @@ class ReconciliationReport extends React.Component {
...
+      const bucket = parsed.hostname.split('.')[0];
+      return {
+        granuleId: d.GranuleUR,
+        filename: path.basename(parsed.pathname),
+        bucket,
+        path: `s3://${bucket}${parsed.pathname}`
+      };
+    });
+
+    return { filesInS3, filesInDynamoDb, filesOnlyInCumulus, filesOnlyInCmr };
```
jennyhliu: I don't know if it's a good idea to put the two `files` report together.
`filesInS3, filesInDynamoDb` is the result of comparing all the files in s3 and dynamodb.
`filesOnlyInCumulus, filesOnlyInCmr` is the result of comparing files for the granules which are in both Cumulus and Cmr.  So `filesOnlyInCumulus` doesn't include the granule files whose granule is not in CMR.  Put this two reports together may cause confusion.
I suggest to list the `filesOnlyInCumulus, filesOnlyInCmr` after the granule report.

app/scripts/components/reconciliation-reports/reconciliation-report.js
```diff
@@ -121,37 +198,75 @@ class ReconciliationReport extends React.Component {
...
-            props={['filename', 'bucket', 'link']}
+            title={'Files only in S3'}
+            tableHeader={tableHeaderS3Files}
+            tableRow={tableRowS3File}
+            tableProps={tablePropsS3File}
+          />
+
+          <ReportTable
+            data={filesOnlyInCumulus}
+            title={'Files only in Cumulus'}
```
jennyhliu: Maybe change to `Granule Files only in Cumulus` and `Granule Files only in CMR` to distinguish it from the cumulus internal file report.

app/scripts/components/reconciliation-reports/report-table.js
```diff
@@ -0,0 +1,44 @@
...
+      tableRow,
+      tableProps
+    } = this.props;
+
+    if (!data || !data.length) {
+      return null;
+    }
+
+    return (
+      <div className='page__section--small'>
```
jennyhliu: Is it possible to add `Hide` and `Show` capability for listing files/granules?   There could be long list of discrepancy.

jennyhliu:
> I also just ran this against my API for my test stack and it worked without issue. However, my files in S3 table had over 11000 rows, which does not seem ideal because you have to scroll for quite a long time to see the rest of the tables.
> 
> However, would this happen for a real deployment? Or is it only happening on my test stack because we have lots of extraneous files in S3?

We are sharing the buckets in sandbox.  However, in real deployment, the list could still be very long.  I think it would be ideal to have `show/hide` effect on the table and also have a count.

PR #568, opened 2019-05-01
CUMULUS-1275 DR Button for Granules

ifestus:
PR for [CUMULUS-1275 - Add a button onto the dashboard for granule DR initiation.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1275)

What was done:
* Added `GRANULE_RECOVER` actions.
* Added a `Recover Granules` bulkAction to the granules overview page.
* Added a `recoveryPath` configuration value to the `app/scripts/config/config.js` file that will allow a user to configure their recovery path (see below for more information there).

The current solutions assumes that the endpoint being pointed at for recovery (`recoveryPath`) exists behind our backend API. If we _do not_ assume this to be the case, then we have to do some extra work with authentication and changing the action object. Making the URL would be definitely do-able.


app/scripts/actions/index.js
```diff
@@ -14,8 +14,11 @@ import log from '../utils/log';
 import * as types from './types';
 
 const CALL_API = types.CALL_API;
-const root = _config.apiRoot;
-const { pageLimit, minCompatibleApiVersion } = _config;
+const {
```
ifestus: this change is just aesthetic

app/scripts/components/granules/overview.js
```diff
@@ -73,7 +74,9 @@ class GranulesOverview extends React.Component {
       }
     };
     const { granules } = this.props;
-    return bulkActions(granules, config);
+    return bulkActions(granules, config).concat(
```
ifestus: This may not be the best way to compose the two functions - I'm open to suggestions.

This way _is_ expandable, though. Any new additions (buttons) can be added by adding them to the `concat` statement after creating an analogous `recoverAction`-esque function.

app/scripts/components/table/list-view.js
```diff
@@ -203,7 +203,7 @@ class List extends React.Component {
         {hasActions ? (
           <div className='form--controls'>
             {this.renderSelectAll()}
-            {bulkActions.map((item, i) => <BatchAsyncCommand key={item.text}
+            {bulkActions.map((item) => <BatchAsyncCommand key={item.text}
```
ifestus: the `i` wasn't used, so I killed it. Feel free to tell me that this is wrong.

app/scripts/components/table/metadata.js
```diff
@@ -9,7 +9,7 @@ class Metadata extends React.Component {
     const { data, accessors } = this.props;
     return (
       <dl className='metadata__details'>
-        {accessors.reduce((acc, meta, i) => {
+        {accessors.reduce((acc, meta) => {
```
ifestus: the `i` wasn't being used so I killed it. Feel free to tell me that I'm wrong.

app/scripts/reducers/config.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+import _config from '../config';
+
+export default (state = { ..._config }) => state;
```
ifestus: Add our configuration values into the `store` for easy access in components.

ifestus: Philosophically consistent with the redux method of `keeping all ui state in the store`

app/scripts/components/granules/overview.js
markdboyd: To me, I think I prefer that the logic controlling what will be rendered (via what gets added to the returned array) here is in this file. For example

```
  generateBulkActions () {
    const actionConfig = {
      execute: {
        options: this.getExecuteOptions(),
        action: this.applyWorkflow
      }
    };
    const { granules, config } = this.props;
    let actions = bulkActions(granules, actionConfig);
    if (config.recoveryPath) {
      actions.concat(recoverAction(granules, actionConfig));
    }
    return actions;
  }
```

test/utils/granules.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+import test from 'ava';
+import { recoverAction } from '../../app/scripts/utils/table-config/granules';
+import { recoverGranule } from '../../app/scripts/actions';
+
+const granules = {
+  recovered: 'recovered'
+};
+
+test('recoverAction returns an empty list when disabled', function (t) {
```
markdboyd: If you make the above recommended change, `recoverAction` itself will have no real logic built into it, so then I would do a more traditional component test where you mock the props/store to the granules overview component and test whether the recovery button is rendered

app/scripts/components/granules/overview.js
ifestus: LGTM :+1:

test/utils/granules.js
ifestus: Good idea. Working on it now.

ifestus: https://github.com/nasa/cumulus-dashboard/pull/568/commits/e2407e9d5aeb25d257b39f14f4ca266d34bc21ef

app/scripts/components/granules/overview.js
ifestus: https://github.com/nasa/cumulus-dashboard/pull/568/commits/e2407e9d5aeb25d257b39f14f4ca266d34bc21ef

app/scripts/components/granules/overview.js
```diff
@@ -161,11 +166,14 @@ GranulesOverview.propTypes = {
   stats: PropTypes.object,
   dispatch: PropTypes.func,
   workflowOptions: PropTypes.array,
-  location: PropTypes.object
+  location: PropTypes.object,
+  config: PropTypes.object
 };
 
+export { GranulesOverview };
```
markdboyd: Why the extra export?

test/components/granules/overview.js
```diff
@@ -0,0 +1,104 @@
...
+  const recoverFilter = (object) => object.text === 'Recover Granule';
+  const recoverActionList = listBulkActions.filter(recoverFilter);
+  t.is(recoverActionList.length, 1);
+
+  const recoverAction = recoverActionList[0];
+  t.truthy(recoverAction);
+  t.is(recoverAction.text, 'Recover Granule');
+});
+
+test('Overview renders bulkAction buttons without recovery button', function (t) {
```
markdboyd:
```suggestion
test('Overview does not generate bulkAction for recovery button', function (t) {
```

test/components/granules/overview.js
```diff
@@ -0,0 +1,104 @@
...
+            key: 'my-key-path/my-name',
+            bucket: 'my-bucket'
+          }
+        ]
+      }
+    }
+  }
+};
+
+test('Overview renders bulkAction buttons with recovery button', function (t) {
```
markdboyd:
```suggestion
test('GranulesOverview generates bulkAction for recovery button', function (t) {
```

test/components/granules/overview.js
```diff
@@ -0,0 +1,104 @@
...
+  const location = { pathname: 'granules' };
+  const config = { recoveryPath: 'recover' };
+  const store = {
+    subscribe: () => {},
+    dispatch: dispatch,
+    getState: () => {}
+  };
+
+  const providerWrapper = shallow(
+    <Provider store={store}>
```
markdboyd: Since the `store` object above returns no state, is this doing anything?

test/components/granules/overview.js
```diff
@@ -0,0 +1,104 @@
...
+      <GranulesOverview
+        granules = {granules}
+        stats = {stats}
+        dispatch = {dispatch}
+        workflowOptions = {workflowOptions}
+        location = {location}
+        config={config}/>
```
markdboyd: So in reality, `config` comes from the store, but in this test, we mock by passing it directly as a prop? That's fine, just making sure that I understand

app/scripts/config/config.js
```diff
@@ -8,6 +8,6 @@ module.exports = {
     }
   },
   apiRoot: process.env.APIROOT || 'https://example.com',
+  // recoveryPath: 'recover',
```
markdboyd: Is there any reason to leave this in commented versus removing it?

app/scripts/actions/index.js
```diff
@@ -312,6 +315,18 @@ export const removeGranule = (granuleId) => ({
   }
 });
 
+export const recoverGranule = (granuleId) => ({
```
markdboyd: Now that we have `config` in the store, it would be cool to re-factor this action (and later the other actions) to read it from the store rather than reading the value read from the file (`_config`). This further isolates our actions to only interact with the store instead of side effects.

Completed untested refactor:

```
export const recoverGranule = (granuleId) => {
    const { config } = getState();
    return dispatch({
      type: types.GRANULE_RECOVER,
      method: 'PUT',
      id: granuleId,
      path: `${config.recoveryPath}/${granuleId}`,
      body: {
        action: 'recoverGranule'
      }
    });
  };
};
```

app/scripts/components/granules/overview.js
ifestus: Currently I'm doing the extra export to avoid having to deal with `connect()(GranulesOverview)`. I can look into doing this without the extra export.

test/components/granules/overview.js
ifestus: 🤔 looking into it

app/scripts/config/config.js
ifestus: not really. This brings up a good point - I should probably add some documentation along with this PR outlining the functionality.

test/components/granules/overview.js
ifestus: If I import `GranulesOverview` class, without the connect, from list-view.js, then I think this `{store}` isn't required. I'm having troubles getting this working without `{store}` passed into the provider - specifically, `dive()` into the `GranulesOverview` throws an error about "not having a store object passed in".

ifestus: This might be something that I can pass in as `dive({ store})`, but at that point it's half-a-dozen to 6.

markdboyd: OK, I'm fine to leave this as is, but I'm a bit confused. If we are wrapping in `<Provider>` with the store, then we should be able to use the exported GranulesOverview with the connect? And so we shouldn't need the extra GranulesOverview export. Or is that not true?

app/scripts/components/granules/overview.js
markdboyd: I just noticed that I'm doing the exact same thing for the reconciliation report tests. So this is 100% fine

test/components/granules/overview.js
markdboyd: I just noticed that I'm doing the exact same thing for the reconciliation report tests. So this is 100% fine

PR #569, opened 2019-05-09
CUMULUS-1259 Dashboard button to kick-off DR workflow for collections

ifestus:
[CUMULUS-1259 Dashboard button to kick-off DR workflow for collections](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1259)


test/components/granules/overview.js
```diff
@@ -65,18 +65,14 @@ test('GranulesOverview generates bulkAction for recovery button', function (t) {
   const recoverFilter = (object) => object.text === 'Recover Granule';
   const recoverActionList = listBulkActions.filter(recoverFilter);
   t.is(recoverActionList.length, 1);
-
-  const recoverAction = recoverActionList[0];
```
ifestus: this is redundant

test/components/granules/overview.js
```diff
@@ -65,18 +65,14 @@ test('GranulesOverview generates bulkAction for recovery button', function (t) {
...
-  t.truthy(recoverAction);
-  t.is(recoverAction.text, 'Recover Granule');
 });
 
 test('GranulesOverview does not generate bulkAction for recovery button', function (t) {
   const dispatch = () => {};
   const workflowOptions = [];
   const stats = { count: 0, histogram: {}, stats: {} };
   const location = { pathname: 'granules' };
-  const config = {};
```
ifestus: figure this better reflects the "null state" of an un-configured `recoveryPath`.

app/scripts/actions/index.js
```diff
@@ -155,6 +155,15 @@ export const deleteCollection = (name, version) => ({
   }
 });
 
+export const recoverCollection = (name, version) => ({
```
markdboyd: Does there need to be an `action` property on this object? I just notice that there is one for `recoverGranule`

app/scripts/utils/table-config/collections.js
```diff
@@ -42,6 +42,19 @@ export const tableSortProps = [
   'timestamp'
 ];
 
+const confirmRecover = (d) => `Recover all granules from ${d} ${strings.collection}(s)?`;
```
markdboyd:
```suggestion
const confirmRecover = (d) => `Recover all collections from ${d} ${strings.collection}(s)?`;
```

markdboyd: Out of curiosity, why is the confirmation message here not just `Recover ${d} ${strings.collections}(s)?`

ifestus: Good question. I think that's a good idea. Ideally the users are going to define their own recovery workflow, so we don't really have insight into what it means to "recover a collection." Probably a good idea to let the user fill in that blank.

app/scripts/actions/index.js
ifestus: Good call. I'm going to add it because there is a possibility of future functionality hitting this same endpoint for different actions.

app/scripts/utils/table-config/collections.js
ifestus: https://github.com/nasa/cumulus-dashboard/pull/569/commits/f73ab300344d599e3f86325cdc989f627a6e5206

app/scripts/actions/index.js
ifestus: https://github.com/nasa/cumulus-dashboard/pull/569/commits/f73ab300344d599e3f86325cdc989f627a6e5206

PR #572, opened 2019-05-24
CUMULUS-799: Homepage distribution metrics

mennovandiermen:
Changes
* Add distribution metrics to homepage


app/scripts/actions/index.js
```diff
@@ -371,6 +371,15 @@ export const getStats = (options) => ({
   }
 });
 
+export const getDistMetrics = () => ({
+  [CALL_API]: {
+    type: types.DIST,
+    method: 'GET',
+    url: url.resolve(root, 'distributionMetrics'),
+    qs: {}
```
markdboyd: Do we need the empty query state?

app/scripts/reducers/dist.js
```diff
@@ -0,0 +1,37 @@
...
+import assignDate from './assign-date';
+
+import {
+  DIST,
+  DIST_INFLIGHT,
+  DIST_ERROR
+} from '../actions/types';
+
+const initialState = {
+  dist: {
```
markdboyd: Is there a reason to have a key of `dist` inside the redux store slice already named `dist`? It seems a bit repetitive.

app/scripts/components/home.js
```diff
@@ -78,6 +81,10 @@ class Home extends React.Component {
       [tally(get(this.props.rules, 'list.meta.count')), 'Ingest Rules', '/rules'],
       [seconds(get(stats.data, 'processingTime.value', nullValue)), 'Average processing Time']
     ];
+    const distStats = [
```
markdboyd: What happens when there is no data for errors or successes and we just have the initial state? Does that need a component test?

fake-api.js
```diff
@@ -195,6 +195,12 @@ app.get('/stats/aggregate', async (req, res, next) => {
   next();
 });
 
+app.get('/distributionMetrics', async (req, res) => {
```
markdboyd: I don't think you need this. We have this line in here later on:

`app.use('/', express.static('test/fake-api/fixtures', { index: 'index.json' }));`

That line will automatically map a request for `distributionMetrics` to a fake API fixture in the `distributionMetrics` folder

test/components/homepage/home.js
```diff
@@ -0,0 +1,40 @@
...
+  const dispatch = () => {};
+
+  const home = shallow(<Home
+    dispatch={dispatch} 
+    dist={dist}
+    granules={granules}
+    stats={stats}
+    />);
+
+  t.is(home.find('#distMetric').length, 2);
```
markdboyd: Can/should we test the specific metric values?

app/scripts/components/home.js
```diff
@@ -110,6 +117,26 @@ class Home extends React.Component {
               </ul>
             </div>
           </section>
+          <section className='page_section'>
+            <div className='row'>
+              <div className='heading__wrapper--border'>
+                  <h2 className='heading--medium heading--shared-content--right'>Distribution Metrics</h2>
+              </div>
+              <ul>
+                {distStats.map(d => {
```
markdboyd: I see that this block is used above for other numbers. Can we copy this render logic into a re-usable method on this component? Or a separate component if it recurs elsewhere

fake-api.js
mennovandiermen: You are right. 👍

test/components/homepage/home.js
mennovandiermen: I don't know that we can with `shallow`, since the list uses a `Link` child component.

app/scripts/actions/index.js
mennovandiermen: Nope 👍

app/scripts/components/home.js
mennovandiermen: 8339f0a

app/scripts/actions/index.js
mennovandiermen: 8339f0a

fake-api.js
mennovandiermen: 8339f0a

app/scripts/components/home.js
mennovandiermen: The distribution metrics simply don't appear when there is no data.

PR #573, opened 2019-05-31
CUMULUS-1248

ifestus:
[CUMULUS-1248 | Integrate PODAAC Disaster Recovery solution with Cumulus Dashboard](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1248)

**ToDo:**
* ~recoverGranule should use applyWorkflow with a workflow specified by the collection meta it belongs to~
* ~recoverCollection should use applyWorkflowToCollection (doesn't exist on the API yet) specified by the collection meta~
* ~Fix previously written unit tests to work with new functionality~
* ~Write new unit tests to better cover functionality~


app/scripts/actions/index.js
```diff
@@ -65,6 +65,7 @@ export const getCollection = (name, version) => ({
     type: types.COLLECTION,
     method: 'GET',
     id: getCollectionId({name, version}),
+    // path: `collections/${name}/${version}`
```
ifestus: removing this

test/actions/recover.js
```diff
@@ -0,0 +1,225 @@
...
+  t.context.defaultConfig = {
+    json: true,
+    resolveWithFullResponse: true,
+    simple: false
+  };
+  store.clearActions();
+});
+
+test.serial('appylyRecoveryWorkflowToCollection fails to acquire collection, dispatches COLLECTION_APPLYWORKFLOW_ERROR', async (t) => {
+  nock('https://example.com')
```
ifestus: these `example.com`s should probably be `localhost:3000`

app/scripts/actions/index.js
markdboyd: I'm confused: in the collections API, the route is specified as:

`router.get('/:name/:version', get);`

which suggests that the commented out path is correct? Perhaps specifying the collection name and version either via the path or as query params work?

app/scripts/actions/index.js
```diff
@@ -303,6 +329,34 @@ export const applyWorkflowToGranule = (granuleId, workflow) => ({
   }
 });
 
+export const applyRecoveryWorkflowToGranule = (granuleId) => {
```
markdboyd: Just trying to follow the logic of what this action is doing:

1. Get the granule by ID
2. Get collection for that granule
3. Get the granule recovery workflow from the collection, if any
4. Dispatch the recovery workflow on the granule or dispatch error

Correct?

app/scripts/actions/index.js
```diff
@@ -303,6 +329,34 @@ export const applyWorkflowToGranule = (granuleId, workflow) => ({
...
+  return (dispatch) => {
+    return dispatch(getGranule(granuleId)).then((granuleResponse) => {
+      const { name, version } = collectionNameVersion(granuleResponse.data.collectionId);
+      return dispatch(
+        getCollection(name, version)
+      ).then((collectionResponse) => {
```
markdboyd: It _might_ be more readable to abstract the logic for getting a collection by granule ID into a separate action that is used here

app/scripts/components/collections/list.js
```diff
@@ -50,9 +51,14 @@ class CollectionList extends React.Component {
   }
 
   generateBulkActions () {
+    const actionConfig = {
+      recover: {
+        action: applyRecoveryWorkflowToCollection
```
markdboyd: I'm sure I'm missing something, but how does this action receive the collection ID?

app/scripts/utils/table-config/collections.js
```diff
@@ -43,14 +43,11 @@ export const tableSortProps = [
...
+export const recoverAction = function (collections, config) {
   return [{
     text: 'Recover',
-    action: (collectionId) => {
-      const { name, version } = collectionNameVersion(collectionId);
-      return recoverCollection(name, version);
-    },
-    state: collections.recovered,
+    action: config.recover.action,
+    state: collections.executed, // this will probably need to be changed
```
markdboyd: What does this do?

app/scripts/components/granules/overview.js
```diff
@@ -89,6 +95,10 @@ class GranulesOverview extends React.Component {
     return applyWorkflowToGranule(granuleId, this.state.workflow);
   }
 
+  applyRecoveryWorkflow (granuleId) {
```
markdboyd: How does `granuleId` get passed in here?

test/actions/recover.js
markdboyd: We're using `nock` to make sure these tests don't go to network, right? So yeah, it seems like they should match whatever URL the components will actually try to request?

test/actions/recover.js
```diff
@@ -0,0 +1,225 @@
...
+    const expectedActionTypes = [
+      'COLLECTION_INFLIGHT',
+      'COLLECTION_ERROR',
+      'COLLECTION_APPLYWORKFLOW_ERROR'
+    ];
+    t.deepEqual(expectedActionTypes, dispatchedActionTypes);
+  });
+});
+
+test.serial('applyRecoveryWorkflowToCollection collection collectionRecoveryWorkflow now configured, dispatches error', async (t) => {
```
markdboyd: Can we re-name this test?

test/actions/recover.js
```diff
@@ -0,0 +1,225 @@
...
+    results: [{
+      meta: {}
+    }]
+  };
+
+  nock('https://example.com')
+    .get(`/collections?name=${name}&version=${version}`)
+    .reply(200, getCollectionResponse);
+
+  return store.dispatch(applyRecoveryWorkflowToCollection(collectionId)).then(() => {
```
markdboyd: Aren't redux thunks fun when they work correctly?

test/actions/recover.js
```diff
@@ -0,0 +1,225 @@
...
+    const expectedActionTypes = [
+      'GRANULE_INFLIGHT',
+      'GRANULE_ERROR',
+      'GRANULE_APPLYWORKFLOW_ERROR'
+    ];
+    t.deepEqual(expectedActionTypes, dispatchedActionTypes);
+  });
+});
+
+test.serial('applyRecoveryWorkflowToGranule related collection does not have recoveryWorkflow configured', async (t) => {
```
markdboyd:
```suggestion
test.serial('applyRecoveryWorkflowToGranule dispatches GRANULE_APPLYWORKFLOW_ERROR when related collection does not have recoveryWorkflow configured', async (t) => {
```

test/actions/recover.js
```diff
@@ -0,0 +1,225 @@
...
+      'GRANULE',
+      'COLLECTION_INFLIGHT',
+      'COLLECTION',
+      'GRANULE_APPLYWORKFLOW_ERROR'
+    ];
+    t.deepEqual(expectedActionTypes, dispatchedActionTypes);
+  });
+});
+
+test.serial('applyRecoveryWorkflowToGranule fails to acquire collection, fails, and dispatches failure', async (t) => {
```
markdboyd:
```suggestion
test.serial('applyRecoveryWorkflowToGranule fails to acquire collection and dispatches GRANULE_APPLYWORKFLOW_ERROR', async (t) => {
```

I don't actually feel strongly about this, but however we name the tests we should be consistent

app/scripts/actions/index.js
ifestus: I kind of gave up on getting this to work in this iteration. I totally agree that the code I have commented here _should_ work (based on the api documentation and local testing), but for some reason it stalls my dashboard when I use it.

I'm going to spend some time trying to debug that here and let you know what I come up with.

test/actions/recover.js
ifestus: My unit tests by default (`npx ava test/actions/recover.js`) are using the `example.com` endpoint. That doesn't seem to be the case for other tests, so I need to figure out what's going on there.

app/scripts/actions/index.js
```diff
@@ -290,6 +279,43 @@ export const reprocessGranule = (granuleId) => ({
...
+});
+
+export const applyRecoveryWorkflowToCollection = (collectionId) => {
+  return (dispatch) => {
+    const { name, version } = collectionNameVersion(collectionId);
+    return dispatch(getCollection(name, version)).then((collectionResponse) => {
+      const collectionRecoveryWorkflow = getProperty(
+        collectionResponse.data.results[0],
+        'meta.collectionRecoveryWorkflow',
+        undefined
```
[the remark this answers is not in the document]
ifestus: Good call - what you describe appears to be the case.

app/scripts/actions/index.js
ifestus: You've got it :100:

app/scripts/actions/index.js
ifestus: Hmm... I think that's a good idea. I'll look into that.

app/scripts/utils/table-config/collections.js
ifestus: This lets the button know that it's state (SUCCESS/FAILED/IN_FLIGHT) is based on `collections.executed`. `collections.executed` is the state variable that is updated by `COLLECTION_APPLYWORKFLOW` actions.

It's worth noting that `granule` works the same way - we set `state: granules.executed` so the button pop-up/status thing takes a queue from the right state variable.

app/scripts/components/granules/overview.js
ifestus: I don't know concretely. It definitely has something to do with https://github.com/nasa/cumulus-dashboard/blob/master/app/scripts/components/table/list-view.js#L206 `BatchAsyncCommand`s... I'll do some more digging.

markdboyd: Hmm okay. Yeah I didn't realize this was just part of how bulk actions work on the dashboard in general. I still don't see where the argument gets passed in, but it's not important to figure out right now

markdboyd: It looks like the code automatically passes the ID associated with the row(s) you selected (I'm not sure how it determines the ID) to the action: https://github.com/nasa/cumulus-dashboard/blob/master/app/scripts/components/form/batch-async-command.js#L79

So yeah, I wouldn't looking into it more

test/actions/recover.js
ifestus: `test/actions/store.js` nocks `https://localhost:5001`, but it's also defining it's `requestAction` such that the `url` matches:
```
const requestAction = {
  type: 'TEST',
  method: 'GET',
  url: 'http://localhost:5001/test-path'
```

The default `apiRoot` configuration for the dashboard points to `example.com` which is why my tests are pointing to `example.com`.

app/scripts/utils/table-config/collections.js
markdboyd: Got it. This seems to be another result of how batch async commands work: https://github.com/nasa/cumulus-dashboard/blob/master/app/scripts/components/form/batch-async-command.js#L39

app/scripts/actions/index.js
ifestus: https://github.com/nasa/cumulus-dashboard/pull/573/commits/8ec2ae5d957121130c164f456bb7e379f892f6d8

app/scripts/actions/index.js
ifestus: https://github.com/nasa/cumulus-dashboard/pull/573/commits/8ec2ae5d957121130c164f456bb7e379f892f6d8

app/scripts/actions/index.js
```diff
@@ -303,6 +328,36 @@ export const applyWorkflowToGranule = (granuleId, workflow) => ({
   }
 });
 
+export const getCollectionByGranuleId = (dispatch, granuleId) => {
```
markdboyd: Every thunk should receive `dispatch` as a parameter, so I don't think we have to pass in `dispatch` this way. I think we can change this to:

```
export const getCollectionByGranuleId = (granuleId) => {
  return (dispatch) => {
    return dispatch(getGranule(granuleId)).then((granuleResponse) => {
      const { name, version } = collectionNameVersion(granuleResponse.data.collectionId);
      return dispatch(getCollection(name, version));
    });
  }
};
```

app/scripts/actions/index.js
```diff
@@ -303,6 +328,36 @@ export const applyWorkflowToGranule = (granuleId, workflow) => ({
...
+  return dispatch(getGranule(granuleId)).then((granuleResponse) => {
+    const { name, version } = collectionNameVersion(granuleResponse.data.collectionId);
+    return dispatch(getCollection(name, version));
+  });
+};
+
+export const applyRecoveryWorkflowToGranule = (granuleId) => {
+  return (dispatch) => {
+    return getCollectionByGranuleId(dispatch, granuleId)
```
markdboyd: Then this would be:

` return dispatch(getCollectionByGranuleId(granuleId))`

ifestus: nice catch :100:

test/actions/recover.js
```diff
@@ -0,0 +1,238 @@
...
+    json: true,
+    resolveWithFullResponse: true,
+    simple: false
+  };
+  store.clearActions();
+});
+
+test.serial('appylyRecoveryWorkflowToCollection fails to acquire collection, dispatches COLLECTION_APPLYWORKFLOW_ERROR', async (t) => {
+  nock('https://example.com')
+    // .get(`/collections/${name}/${version}`)
```
markdboyd: Can we remove these lines?

PR #578, opened 2019-06-06
Cumulus-1192 Added Granule csv download 

kkelly51:
Cumulus-1192 
Added CSV download button to granules page


app/scripts/reducers/granule-csv.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+import {
+  GRANULE_CSV,
+  GRANULE_CSV_INFLIGHT,
+  GRANULE_CSV_ERROR
+} from '../actions/types';
+
+export const initialState = {
+  list: {
```
markdboyd: Reading your reducer below, it seems like the only properties you assign to the state are `data`, `inflight`, and `error`:

`{ data: action.data, inflight: false, error: null }`

So it seems like the rest of the properties on the initialState here aren't necessary? Also, you may want to make sure that you have initial values for the properties that you are using

app/scripts/components/granules/overview.js
```diff
@@ -115,10 +118,20 @@ class GranulesOverview extends React.Component {
     return <Overview items={overview} inflight={false} />;
   }
 
+  csvDownloadSection (fileData) {
+    if (fileData) {
+      const data = new Blob([fileData], {type: 'text/csv'});
+      const url = window.URL.createObjectURL(data);
+
+      return (<a className='csv__download' id='download_link' download='granules.csv' href={url}>Download Granule List</a>);
```
markdboyd: The ticket mentions integration tests. Is the plan to add a Cypress test that shows at least that the link exists when CSV data comes back from the API? It would also be interesting to test actually clicking the download link to see if you get the right behavior, but that may be more challenging

app/scripts/actions/index.js
```diff
@@ -398,6 +398,15 @@ export const clearGranulesSearch = () => ({ type: types.CLEAR_GRANULES_SEARCH })
 export const filterGranules = (param) => ({ type: types.FILTER_GRANULES, param: param });
 export const clearGranulesFilter = (paramKey) => ({ type: types.CLEAR_GRANULES_FILTER, paramKey: paramKey });
 
+export const getGranuleCSV = (options) => ({
+  [CALL_API]: {
+    type: types.GRANULE_CSV,
+    method: 'GET',
+    url: url.resolve(root, 'granule-csv'),
+    qs: Object.assign({ limit: pageLimit }, options)
```
markdboyd: Does this API endpoint actually respect any query string or can this be removed?

app/scripts/components/granules/overview.js
```diff
@@ -115,10 +118,20 @@ class GranulesOverview extends React.Component {
     return <Overview items={overview} inflight={false} />;
   }
 
+  csvDownloadSection (fileData) {
+    if (fileData) {
```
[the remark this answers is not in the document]
kkelly51: Yeah you're right, I like this better.

app/scripts/actions/index.js
kkelly51: removed.

app/scripts/actions/index.js
```diff
@@ -398,6 +398,15 @@ export const clearGranulesSearch = () => ({ type: types.CLEAR_GRANULES_SEARCH })
 export const filterGranules = (param) => ({ type: types.FILTER_GRANULES, param: param });
 export const clearGranulesFilter = (paramKey) => ({ type: types.CLEAR_GRANULES_FILTER, paramKey: paramKey });
 
+export const getGranuleCSV = (options) => ({
+  [CALL_API]: {
+    type: types.GRANULE_CSV,
+    method: 'GET',
+    url: url.resolve(root, 'granule-csv'),
+    qs: Object.assign({}, options)
```
markdboyd: You don't need query state at all, right?

app/scripts/components/granules/overview.js
```diff
@@ -115,10 +118,21 @@ class GranulesOverview extends React.Component {
...
+    return (<a className='csv__download' id='download_link' download='granules.csv' href={url}>Download Granule List</a>);
+  }
+
   render () {
-    const { stats, granules } = this.props;
+    const { stats, granules, granuleCSV } = this.props;
     const { list, dropdowns } = granules;
     const { count, queriedAt } = list.meta;
+    const { data } = granuleCSV;
+    console.log(data);
```
markdboyd: We can remove this

app/scripts/actions/index.js
kkelly51: nope, will remove

app/scripts/components/granules/overview.js
kkelly51: good catch

app/scripts/reducers/granule-csv.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+import {
+  GRANULE_CSV,
+  GRANULE_CSV_INFLIGHT,
+  GRANULE_CSV_ERROR
+} from '../actions/types';
+
+export const initialState = {
+  data: {},
```
markdboyd: When data is received, it's just a string of text, correct? If so, I think it makes sense to default to `null`. Then you should also not need the `Object.entries(fileData).length === 0` above

kkelly51: fixed

PR #579, opened 2019-07-18
Cumulus 1337 link to kibana

flamingbear:
Adds two new ButtonListSections to the dashboard homepage.
`Distribution Successes` and `Distribution Errors`

These lists appear when the dashboard is configured to get data from metrics' teams ELK stack using Kibana and Elasticsearch.  The buttons display the counts of each type of action (success or failure) and the buttons have a link to the kibana instance showing the discovery search.

- [ ] Add A "pre-release" PR to remind us to update the verions needed in the README/CHANGELOG and the minCompatibleApiVersion.

Note: This links to documentation in cumulus that hasn't been written/approved yet.


app/scripts/components/home.js
```diff
@@ -41,6 +54,14 @@ class Home extends React.Component {
...
       this.query();
     }, updateInterval, true);
+    const {dispatch} = this.props;
+    dispatch(getCumulusInstanceMetadata())
+      .then(() => {
+        dispatch(getDistApiGatewayMetrics(this.props.cumulusInstance));
+        dispatch(getDistApiLambdaMetrics(this.props.cumulusInstance));
+        dispatch(getDistS3AccessMetrics(this.props.cumulusInstance));
+      }
+    );
```
flamingbear: I did this to try to make sure the stackName was populated before dispatching the calls to ES, but it seems that sometimes I see the dashboard with zero values across the board before the  buttons update with the correct values.  If there's a better way, please tell me.

cors-proxy.js
```diff
@@ -0,0 +1,18 @@
+// server to proxy CORS requests for tunneled resources
+// via localhost port forwarding during development.
+
+// Listen on a specific host via the CPHOST environment variable
+var host = process.env.CPHOST || '127.0.0.1';
```
laurenfrederick: Where would these environment variables get set? Do we have any documentation on when they should be set and what they should be set to?

flamingbear: No, I basically added them for someone who didn't want to use the default localhost ip and port that I selected (which is what basically anyone would want).  The would be set when you run the command if you wanted something different.

```
CPHOST=#LocalIP PORT=#portnumber node cors-proxy.js
```

PR #583, opened 2019-09-06
Cumulus 1427 all granules show as errors

chuckwondo:
Fully initialize List state in constructor

The state of a List component was not fully initialized by the constructor and relied upon the `componentDidMount` method to finish state initialization, which was bad.  Not only should state not be modified in `componentDidMount`, but the part of the state that was not initialized in the constructor was the query criteria for fetching granules.

This meant that _all_ granules were fetched and initially populated in the List component, which caused the error granules list on the home page to list _all_ granules, not only error granules.

The state is now fully initialized in the List constructor.

In addition, this commit includes a bit of additional cleanup and refactoring throughout, including potential reduction of calls to `this.setState` in `componentDidUpdate`, which should probably also not exist.  This method should likely be revisited later to determine how all calls to `this.setState` might be removed from it.

Added a test and updated ignored files to include IntelliJ project files and yarn error log.

Fixes CUMULUS-1427


app/scripts/components/table/list-view.js
```diff
@@ -21,63 +26,61 @@ class List extends React.Component {
...
-        validParams[key] = value;
-      }
-    }
-
-    if (JSON.stringify(validParams) !== JSON.stringify(this.state.params)) {
-      this.setState({ params: validParams }, () => this.setState({ // eslint-disable-line react/no-did-update-set-state
-        queryConfig: this.config() }));
-    }
+    // Remove parameters with null or undefined values
+    const params = omitBy(list.params, isNil);
```
markdboyd: Much cleaner

app/scripts/components/table/list-view.js
```diff
@@ -21,63 +26,61 @@ class List extends React.Component {
...
-    const validParams = {};
-    for (let key in params) {
-      let value = params[key];
-      if (!undef(value) && value !== null) {
-        validParams[key] = value;
-      }
-    }
-
-    if (JSON.stringify(validParams) !== JSON.stringify(this.state.params)) {
-      this.setState({ params: validParams }, () => this.setState({ // eslint-disable-line react/no-did-update-set-state
```
markdboyd: Hmm, why was it waiting to update the queryConfig state until *after* the params state was updated? Is it bad that it doesn't work that way now?

app/scripts/components/table/list-view.js
markdboyd: Much cleaner

app/scripts/components/table/list-view.js
markdboyd: Hmm, why was it waiting to update the queryConfig state until after the params state was updated? Is it bad that it doesn't work that way now?

app/scripts/components/table/list-view.js
```diff
@@ -21,63 +26,61 @@ class List extends React.Component {
...
+    const params = omitBy(list.params, isNil);
 
-    if (sortIdx !== this.state.sortIdx) {
-      this.setState({ sortIdx: sortIdx }); // eslint-disable-line react/no-did-update-set-state
+    if (
+      !isEqual(query, prevProps.query) ||
+      !isEqual(params, this.state.params) ||
+      sortIdx !== this.state.sortIdx
+    ) {
+      const queryConfig = this.getQueryConfig({ params }, query);
```
markdboyd: So previously the queryConfig state was only updated if that specific prop had changed, but now it is updated under all these other conditions as well. Is that a problem?

app/scripts/components/table/list-view.js
```diff
@@ -21,63 +26,61 @@ class List extends React.Component {
...
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ params, queryConfig, sortIdx });
     }
   }
 
   queryNewPage (page) {
-    this.setState({ page });
     this.setState({
-      queryConfig: this.config({ page }),
+      page,
```
markdboyd: 👍

app/scripts/components/table/list-view.js
```diff
@@ -90,30 +93,25 @@ class List extends React.Component {
...
-      if (typeof rowId === 'function') {
-        selected = data.map(rowId);
-      } else {
-        selected = data.map(d => d[this.props.rowId]);
-      }
+    const { rowId, list: { data } } = this.props;
+
+    if (!isEmpty(data)) {
+      const rowIdFn = isFunction(rowId) ? rowId : row => row[rowId];
+      const allSelected = this.state.selected.length === data.length;
```
markdboyd: Definitely cleaner. I'm a bit confused why when `allSelected` is `true`, then we set `selected` to `[]`, but that seems to be what it was doing before

app/scripts/components/table/list-view.js
```diff
@@ -90,30 +93,25 @@ class List extends React.Component {
...
     const { selected } = this.state;
-    if (selected.indexOf(id) === -1) {
-      this.setState({ selected: selected.concat([id]) });
-    } else {
-      this.setState({ selected: selected.filter(d => d !== id) });
-    }
+
+    this.setState({
+      selected: selected.includes(id)
+        ? selected.filter(anId => anId !== id)
```
markdboyd: 👍 for specifically named variables

app/scripts/components/table/list-view.js
```diff
@@ -126,42 +124,39 @@ class List extends React.Component {
...
+      ...config,
+      ...query
+    }, isEmpty);
   }
 
   renderSelectAll () {
-    const { list } = this.props;
-    const allChecked = this.state.selected.length === list.data.length && list.data.length;
+    const { list: { data } } = this.props;
+    const allChecked = !isEmpty(data) && this.state.selected.length === data.length;
```
markdboyd: Better evaluation order 👍

app/scripts/components/table/list-view.js
```diff
@@ -176,9 +171,14 @@ class List extends React.Component {
       tableSortProps,
       bulkActions,
       rowId,
-      list
+      list,
+      list: {
```
markdboyd: You can de-structure the same property twice? In this case allowing you to also de-structure nested properties. I don't know that I think it's clearer, but I don't have any issue with it

app/scripts/components/table/list-view.js
```diff
@@ -189,7 +189,7 @@ class List extends React.Component {
       bulkActionError
     } = this.state;
     const primaryIdx = 0;
-    const hasActions = Array.isArray(bulkActions) && bulkActions.length;
+    const hasActions = Array.isArray(bulkActions) && bulkActions.length > 0;
```
markdboyd: Honest question: What is the difference between testing `bulkActions.length` and `bulkActions.length > 0?

app/scripts/components/table/list-view.js
```diff
@@ -231,11 +233,16 @@ class List extends React.Component {
           order={order}
           changeSortProps={this.queryNewSort}
           onSelect={this.updateSelection}
-          canSelect={!!hasActions}
+          canSelect={hasActions}
```
markdboyd: I assume that the double bang was to cast this as a boolean? Is it already a boolean? I guess it must be or else the props validation on SortableTable would fail when the prop is passed

test/components/table/list-view.js
```diff
@@ -0,0 +1,57 @@
...
+import {
+  errorTableHeader,
+  errorTableRow, errorTableSortProps
+} from '../../../app/scripts/utils/table-config/granules';
+
+configure({ adapter: new Adapter() });
+
+test('list-view should properly initialize timer config prop', async (t) => {
+  const dispatch = () => Promise.resolve();
+  const query = { q: '_exists_:error AND status:failed' };
```
markdboyd: Nice! So this proves what we care about - that when we pass in a query (really any query), it should be properly initialized by the component without any lifecycle methods

test/components/table/list-view.js
```diff
@@ -0,0 +1,57 @@
...
+      query={query}
+    />,
+    {
+      // We must disable lifecycle methods so that when we find the Timer its
+      // properties will still be set to their initial values.  Otherwise, a
+      // lifecycle method might cause the properties to change and we will not
+      // be able to check their initial values, which is what we want to do.
+      disableLifecycleMethods: true,
+      context: {
+        store: {
```
markdboyd: So is it only working with you when you provide the store this way instead of via the `Provider`? If so, I don't care or want it to change, it's just weird

test/components/table/list-view.js
```diff
@@ -0,0 +1,57 @@
...
+          subscribe: () => void 0,
+          dispatch: () => void 0,
+          getState: () => void 0
+        }
+      }
+    }
+  );
+
+  const timerWrapper = listWrapper.dive().find(Timer);
```
markdboyd: Interesting that you are doing a find passing in the component itself and not its name. I assume that proved necessary?

test/components/table/list-view.js
```diff
@@ -0,0 +1,57 @@
...
+
+  t.truthy(timerWrapper.exists(), 'Timer not found');
+  // Is the Timer's query configuration properly initialized via the
```
markdboyd: Why is this phrased as a question? The question is what the test is proving, correct?

app/scripts/components/table/list-view.js
chuckwondo: Actually, it does still work that way in essence. However, with the refactoring, only 1 call to `setState` is necessary instead of two. The incremental state object is combined by the new lines of code so that only 1 call is necessary. See the new lines 65-68.

app/scripts/components/table/list-view.js
chuckwondo: Not a problem since if any 1 of the conditions is true, a call to `setState` is made anyway. If part of the state didn't actually change, then it doesn't matter. For example, if `sortIdx` changed, but `queryConfig` didn't, the call to `setState` is required anyway. Including a value for `queryConfig` that is the same as the existing state value won't matter.

By combining everything into a single call to `setState` we never have to make multiple calls to it when multiple conditions are true.

app/scripts/components/table/list-view.js
chuckwondo: That's because if everything is already selected and "select all" is clicked, we want to de-select all (i.e., select none).

app/scripts/components/table/list-view.js
chuckwondo: Yep, no need for a separate destructuring assignment. It's arguably slightly more clear because of proximity. With the previous separate destructuring assignment, the reader must search for where `list` is declared. With this syntax, it appears on the line above, so there's no need to "hunt" for the declaration. Yes, very minor, but I found it helpful.

app/scripts/components/table/list-view.js
chuckwondo: It's a matter of a specific "truthy" value. Without the addition of `> 0`, the value assigned to `hasActions` is the actual length of `bulkActions`, not a boolean value. The problem with that is that when you are then forced to coerce it to a boolean value when you want an actual boolean value, which is what was being done on the old line 234: `!!hasActions`. Now that `> 0` is added, `hasActions` is a true boolean value and I can change `!!hasActions` to `hasActions` (which I did), which is arguably more intuitive and readable.

app/scripts/components/table/list-view.js
chuckwondo: See my reply above about the addition of `> 0`. Regarding validation, using `!!` is required to coerce `hasActions` to a boolean value (specifically to `true` when `hasActions` is _truthy_; `false` otherwise) because without the addition of `> 0` above, it is an integer value (specifically the length of the `bulkActions` array).

test/components/table/list-view.js
chuckwondo: Yeah, I don't know why using the Provider with a store didn't work. In that case, the Timer couldn't be found. I could have added the store as a property on the List component, but chose to add it to the options context instead so that the List component appears just like it does in our usage of it.

test/components/table/list-view.js
chuckwondo: Right

test/components/table/list-view.js
chuckwondo: Precisely. I just wish we had found the `disableLifecycleMethods` option for the `shallow` method sooner :-)

test/components/table/list-view.js
chuckwondo: Not necessary, actually. I could have used the string `'Timer'`, but when I can use specific references rather than strings, I prefer to use them. It enables code analysis to find invalid names prior to runtime. With strings, wrong names aren't found until runtime.

app/scripts/components/table/list-view.js
markdboyd: Good catch and explanation 👍

PR #585, opened 2019-10-01
CUMULUS-1308 handle deleted fields via api

chuckwondo:
Updated dashboard to pass full Provider object to API when updating Provider via HTTP PUT.

This change is to avoid breaking the dashboard when the updated Cumulus API is released and used with the dashboard. The API change that would otherwise break the dashboard is also under CUMULUS-1308, and involves changing PUT operations for Collections, Providers, and Rules to expect full objects in the payload, not partial objects.

Also fixed a number of minor bugs.

Fixes CUMULUS-1308


cypress/integration/collections_spec.js
```diff
@@ -62,40 +62,42 @@ describe('Dashboard Collections Page', () => {
...
-      cy.get('form .ace_content')
-        .within(() => {
-          cy.contains(`"name": "${name}"`);
-          cy.contains(`"version": "${version}"`);
-          cy.contains(`"dataType": "${name}"`);
-          cy.contains(`"duplicateHandling": "${duplicateHandling}"`);
-        });
-
-      // verify the new collection is added to the collections list
-      cy.contains('a', 'Back to Collections').click();
```
markdboyd: I don't think we want to lose this test assertion proving that the collection shows up in the collections list

chuckwondo: I'm on the fence with that suggestion because that's arguably not the responsibility of this particular test, which is simply to make sure that when the user adds a new the collection it is persisted and the user is taken to the page for the new collection itself.

Navigating back to the collections list page and then testing that the collections list page includes the newly added collection (or more generally, the page lists all persisted collections) is arguably the responsibility of a different test, and the test "should display expected MMT Links for collections list" does indeed test that case by expecting a list of all known collections (in the context of the test) to appear.

Given that the test here proves the collection is persisted, and the "MMT Links" test proves that all persisted collections are listed, I find the lines I've removed here are unnecessary.

However, if you feel strongly about this, I'm not adamantly opposed to it.

markdboyd: Hmm, in theory the tests are slightly different I think? The test in "MMT links" just proves that the correct collections show up for what already is persisted in the API. This test is proving that a collection is not only added correctly so that it is returned by a direct /GET request, but persisted in such a way that it shows up in **listing** /GET requests. This seems critical given that the actual persistence layer behind the listing requests (Elasticsearch) is different than behind a single GET request (Dynamo). I know that our dashboard code tests should be ignorant of that distinction, but it feels like they should test all of the expected UI changes associated with an operation such as "add a collection", which would have a nice side effect of also testing that the persistence to both layers is working properly. So I would vote in favor of including this test

chuckwondo: Ah, I wasn't aware of that distinction, so I agree that this makes sense, and I will update per your request.

chuckwondo: @markdboyd I restored the extra check, along with leveraging CMR response mocking, which has resolved the flakiness of the "add collection" test.

app/scripts/reducers/cumulus-instance.js
```diff
@@ -1,25 +1,25 @@
...
+
+  if (action.type === ADD_INSTANCE_META) {
+    state = {
+      ...state,
+      ...fromPairs([
+        ['cmrEnvironment', get(data, 'cmr.environment')],
+        ['cmrProvider', get(data, 'cmr.provider')],
+        ['stackName', get(data, 'cumulus.stackName')]
+      ].filter(([_, value]) => value))
+    };
```
markdboyd: This isn't blocking, but usually reducers have to handle multiple actions, hence the switch statements. Of course you can achieve the same thing with if statements, but it's sort of idiomatic to use switch and that's how the rest of the code in the dashboard is written. I'd prefer the consistency for the next person who might need to handle another action here in this reducer, but it's not critical

chuckwondo: Yep, makes sense. Since there was only 1 condition, I made it an `if` because the IDE was telling me it was silly to use `switch` in this case, and I like to eliminate as many warnings in my IDE as I can.

PR #586, opened 2019-10-04
Update saml login API starting point.

flamingbear:
Adds optional configuration to use SAML protected Cumulus via Launchpad.

Adds REDUX Developer Tools to the build for debugging.

### DIFF
--- a/app/scripts/config/config.js
+++ b/app/scripts/config/config.js
@@ -8,6 +8,7 @@ module.exports = {
     }
   },
   apiRoot: process.env.APIROOT || 'https://example.com',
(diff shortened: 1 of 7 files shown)


app/scripts/config/config.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
     }
   },
   apiRoot: process.env.APIROOT || 'https://example.com',
+  oauthMethod: process.env.AUTH_METHOD || 'earthdata',
```
markdboyd: This should be added to the README

markdboyd: Also needs to be supported in `bin/build_in_docker.sh`

PR #588, opened 2019-10-08
Cumulus 1456 mhs suggest

flamingbear:
I think this is clearer what is happening. One side effect is that we get really long (and useless) errors now when you can't connect to the API, but that could be fixed.
Screen Shot 2019-10-08 at 5 10 04 PM

But you can see the cumulus api errors now in the console.

Screen Shot 2019-10-08 at 5 07 29 PM

It also fixes random dashboard version nit-pick
Screen Shot 2019-10-08 at 5 09 27 PM

Fixes catch

### DIFF
--- a/app/scripts/middleware/request.js
+++ b/app/scripts/middleware/request.js
@@ -11,12 +11,6 @@ import log from '../utils/log';
 import { isValidApiRequestAction } from './validate';
 
 const handleError = ({ id, type, error, requestAction }, next) => {
-  console.group(['handleError']);
-  console.log(`id: ${id}`);
-  console.log(`type: ${type}`);
-  console.log(`error: ${error}`);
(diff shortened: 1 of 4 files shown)


app/scripts/middleware/request.js
```diff
@@ -11,12 +11,6 @@ import log from '../utils/log';
 import { isValidApiRequestAction } from './validate';
 
 const handleError = ({ id, type, error, requestAction }, next) => {
-  console.group(['handleError']);
-  console.log(`id: ${id}`);
-  console.log(`type: ${type}`);
-  console.log(`error: ${error}`);
-  console.log(`requestAction: ${JSON.stringify(requestAction, null, 2)}`);
-  console.groupEnd();
```
flamingbear: Actually maybe this should stay, but remove the `null, 2)` on the stringify.

PR #594, opened 2019-10-18
Cumulus 123 mhs

flamingbear:
Dashboard filters and searches can have their state in the URL so that they are shareable.


app/scripts/utils/url-helper.js
```diff
@@ -0,0 +1,40 @@
...
+  if (value.length) {
+    nextQuery = { ...nextQuery, [paramKey]: value };
+  } else {
+    delete nextQuery[paramKey];
+  }
+  if (location.query[paramKey] !== nextQuery[paramKey]) {
+    location.query = nextQuery;
+    router.push(location);
+  }
+}
```
flamingbear: These could probably use some tests....

app/scripts/components/table/list-view.js
```diff
@@ -144,7 +143,7 @@ class List extends React.Component {
       page: this.state.page,
       order: this.state.order,
       sort_by: this.getSortProp(this.state.sortIdx),
-      params: this.state.params,
+      ...this.state.params,
```
markdboyd: Are these equivalent? The former operation would place the contents of `this.state.params` on the `params` property, whereas the spread operation will put the contents of `this.state.params` directly on the object.

app/scripts/utils/status.js
```diff
@@ -6,22 +6,9 @@ const status = {
 };
 export default status;
 
-export const queryStatus = [
-  'running',
-  'completed',
-  'failed'
-];
-
```
markdboyd: Why is this getting deleted? Unused?

cypress/integration/granules_spec.js
```diff
@@ -107,16 +109,52 @@ describe('Dashboard Granules Page', () => {
...
+      cy.get('#form-Status-status > div > input').as('status-input');
+      cy.get('@status-input').should('have.value', 'Completed');
+    });
+
+    it('Should update dropdown with status value if query params are not a valid status.', () => {
+      cy.visit('/#/granules?status=notanoption');
+      cy.get('#form-Status-status > div > input').as('status-input');
+      cy.get('@status-input').should('have.value', 'notanoption');
+    });
+
```
markdboyd: Hmm, is this a behavior that we want? That entering arbitrary values in the URL actually show up in the component?

cypress/integration/granules_spec.js
```diff
@@ -107,16 +109,52 @@ describe('Dashboard Granules Page', () => {
...
+    it('Should update URL when dropdown filters are activated.', () => {
+      visitGranulePage();
+      cy.get('#form-Status-status > div > input').as('status-input');
+      cy.get('@status-input').click().type('{downarrow}').type('{enter}');
+      cy.url().should('include', '?status=running');
```
markdboyd: I don't love that this test assumes an order, but don't see any alternative

cypress/integration/granules_spec.js
```diff
@@ -107,16 +109,52 @@ describe('Dashboard Granules Page', () => {
...
+    });
+
+    it('Should show Search and Dropdown filters in URL.', () => {
+      visitGranulePage();
+      cy.get('.search').as('search');
+      cy.get('@search').click().type('L2');
+      cy.get('#form-Status-status > div > input').as('status-input');
+      cy.get('@status-input').click().type('{downarrow}{downarrow}{enter}');
+      cy.url().should('include', 'search=L2');
+      cy.url().should('include', 'status=completed');
```
markdboyd: I think you could write this as

`cy.url().should('include', 'search=L2').and('include', 'status=completed')`

.eslintrc
```diff
@@ -6,7 +6,8 @@
       "mocha": true
     },
     "plugins": [
-      "react"
+      "react",
+      "react-hooks"
```
markdboyd: Are we using React hooks anywhere in the code now?

app/scripts/components/form/dropdown.js
```diff
@@ -21,39 +23,68 @@ function renderMenu (items, value, style) {
...
-      value: ''
+      value: initialValue
     };
     this.onSelect = this.onSelect.bind(this);
     this.onChange = this.onChange.bind(this);
   }
 
+  componentDidUpdate (prevProps, prevState, snapshot) {
+    const {location, paramKey, dispatch, action, options} = this.props;
+    if (location.query[paramKey] !== prevProps.location.query[paramKey]) {
```
markdboyd: Hmm, is `location.query` always defined, even on initial component load? I'm surprised that this never errors.

app/scripts/components/form/dropdown.js
```diff
@@ -21,39 +23,68 @@ function renderMenu (items, value, style) {
...
+      let value = location.query[paramKey];
+      dispatch(action({ key: paramKey, value }));
+      if (value) value = statusToLabel(options, value);
+      this.setState({value}); // eslint-disable-line react/no-did-update-set-state
```
markdboyd: You only want to do `this.setState` if you got a value from `statusToLabel`, correct? So this should this statement be inside the `if` block?

app/scripts/utils/url-helper.js
markdboyd: Yeah, I wonder if there's a package providing middleware that would handle this for us. But we can discuss this offline.

test/reducers/test-removeDeleted.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
markdboyd: Very nice. Love to see additional tests 👍

app/scripts/components/form/search.js
```diff
@@ -48,20 +64,27 @@ class Search extends React.Component {
 
   render () {
     return (
-      <div className='filter__item'>
-        <form className='search__wrapper form-group__element' onSubmit={this.submit} value={this.state.value}>
-          <input className='search' type='search' onChange={this.complete}/>
-          <span className='search__icon'/>
+        <div className='filter__item'>
```
markdboyd: Why remove the indentation on this line and line 72?

.eslintrc
flamingbear: I updated to a version that will take hooks, and included the hook linter to be prepared.

app/scripts/utils/url-helper.js
flamingbear: def willing to entertain different ideas.

app/scripts/components/table/list-view.js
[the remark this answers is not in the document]
flamingbear: No, they're not equivalent and the added spread fixes a bug where filters weren't working.

app/scripts/utils/status.js
flamingbear: I couldn't find any use of those variables in our code, so I just deleted because they were confusing me.  If you can think of a reason to keep them, please let me know.

cypress/integration/granules_spec.js
flamingbear: It would be the equivalent if you typed something in the dropdown box that didn't match any of the configured options.  I don't know if we want to prevent that or not?

Ok, I just actually tested that. and if you type a `z` in the dropdown and it doesn't match anything, nothing is kicked off, and if you hit return (select), the url pushes a ? before the hash so we get `http://localhost:3000/?#/granules` . That doesn't seem right.  

I'll have to figure out how to handle this with search params differently from dropdowns.  But Ok I can see this might not be what we want.

To clarify:  you think that if you browse to a url that has a invalid status key, the dropdown corresponding to that paramKey should clear itself, and remove that from the query?

so basically L133 should be `      cy.get('@status-input').should.not('have.value', 'notanoption');`

cypress/integration/granules_spec.js
flamingbear: Let me try something.

app/scripts/components/form/dropdown.js
flamingbear: By creating the router history with `useRouterHistory(createHashHistory)()`, there will always be a query object on a location.   It may be empty, but then these are undefined and not errors.

app/scripts/utils/status.js
markdboyd: Nope, that's fine with me

app/scripts/components/form/dropdown.js
flamingbear: No, because we always want to set the state's value no matter what the user has typed, but if the user types running, we will correct it to Running.

And if the user deleted everything in the box, we want to set the value to the empty string.

cypress/integration/granules_spec.js
markdboyd: I think if the URL has an invalid status key, basically nothing should happen, so whatever logic initially sets the value should short-circuit if the key is invalid.

We probably also don't want the case you described where typing a invalid value in a dropdown and hitting enter affects the URL

app/scripts/components/form/search.js
flamingbear: emacs. autoformating problems I would guess.

app/scripts/components/form/dropdown.js
markdboyd: Ah, got it

test/reducers/test-removeDeleted.js
flamingbear: Well, I had an error in that and I had no idea why, so had to figure out what it was doing.

cypress/integration/granules_spec.js
flamingbear: What should the state of the URL be after discovering a bad value?
say `?status=badvalue`, should that be removed to match the dropdown value that is also blank?

In the case of the problem with the URL updating badly, that's an existing bug that I filed a ticket for https://bugs.earthdata.nasa.gov/browse/CUMULUS-1621, but I'm hoping that it will go away in an update to react router or an update to a browser history instead of a hash history.

app/scripts/components/form/search.js
flamingbear: e382c25

cypress/integration/granules_spec.js
flamingbear: 5e5ef70

cypress/integration/granules_spec.js
flamingbear: f16adad

PR #597, opened 2019-11-01
Cumulus 1461

jjmccoy:
Granules error table, granule updates, and general home page stylings.


jjmccoy:
Also, did the approved file restructure of the app and re-routed everything in the configs and gulpfile.

app/humans.txt
```diff
@@ -14,3 +14,4 @@ Ali Felskia
 Alireza
```
kkelly51: Should we change this whole file to just the listing of Cumulus OX people because none of those other people work on this anymore? And delete the Dev Seed name

app/src/css/_base.scss
```diff
@@ -265,15 +290,32 @@ a:active {
 }
 
 .page__section {
-  margin-bottom: 7em;
+  margin-bottom: 3em;
   @extend .clearfix;
   &:first-of-type {
-    margin-bottom: 4em;
+    margin-bottom: 0em; /* This is temporary until we had the datepicker component */
```
kkelly51: 👍

app/humans.txt
jjmccoy: That would be up to Lauren since she is the Product Owner. She may want to update this. I just added the Cumulus OX team to it because it was missing.

kkelly51: And let's update this while we're at it

jjmccoy:
**PR Feedback Answers**

- Table Options: This table section affects all tables throughout the dashboard. It's the 'table options' and this was what Chuck and I were speaking about over a week ago that the configs are for static pages and this is dynamic and it's out of scope for this ticket. The option allows the user to select how many rows that show in the table at a time. We will need 2 tickets for the 'table options' component - one for building the configs and interactions and another one for styling.

- Granule Failed Color Indicator: We can leave as black for right now but I did remove the 'Granule Failed' always red coloring even when '0' per request in ticket [CUMULUS-757](https://bugs.earthdata.nasa.gov/browse/CUMULUS-757). We have to figure out the threshold calculation for what green, yellow, and red will be in this section. Green would be 0 but what is the formula for yellow and red. That has to be something to be revisited in 1469.

- Execution Page: Please see ticket [CUMULUS-1461](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1461)

PR #602, opened 2019-11-11
Cumulus 1308 lauren

laurenfrederick:
Enabling/disabling was throwing schema validation errors, so fixed in this PR

### DIFF
--- a/app/src/js/actions/index.js
+++ b/app/src/js/actions/index.js
@@ -779,18 +779,28 @@ export const deleteRule = (ruleName) => ({
   }
 });
 
-export const enableRule = (payload) => ({
-  [CALL_API]: {
-    id: payload.name,
-    type: types.RULE_ENABLE,
-    method: 'PUT',
-    path: `rules/${payload.name}`,
-    body: {
-      ...payload,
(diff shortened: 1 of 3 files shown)


app/src/js/actions/index.js
```diff
@@ -778,31 +779,51 @@ export const deleteRule = (ruleName) => ({
...
-    body: {
-      ...payload,
-      state: 'ENABLED'
-    }
+export const enableRule = (payload) => {
+  const rule = clonedeep(payload);
+  delete rule['queriedAt'];
+  delete rule['timestamp'];
+
+  if (!rule.rule.value) {
```
laurenfrederick: Need to fix this on the API side too, since if you remove it it'll put it back in as 
```
rule: { value: null }
```
and null is not ok since we're not coercing the types and it has to be a string.
But we can get a dashboard release out pretty quickly with these fixes.

chuckwondo: I'm not following you here. Enabling the rule is only changing the value of the `state` property on the rule. Can you clarify as to why you believe this code change is necessary?

PR #608, opened 2019-11-18
1507 Collection - Individual - Granules 

kkelly51:
- added new routes in collections category for granule statuses 
- added granule status option to sidebar 
- added granule search to new granule status pages


kkelly51:
Sidebar styling will be done as part of 1508 in sprint 5

PR #611, opened 2019-11-27
CUMULUS-1670: Fix provider update

markdboyd:
Fixes https://bugs.earthdata.nasa.gov/browse/CUMULUS-1670

I have submitted this PR as a hotfix against the `master` branch so we can get this fix out to users of the dashboard ASAP and not necessarily have to merge everything from `develop` into `master`

I have tested this against running the full Cumulus API locally, but still want to double check against my deployed Cumulus API.

### DIFF
--- a/app/src/js/actions/index.js
+++ b/app/src/js/actions/index.js
@@ -781,8 +781,6 @@ export const deleteRule = (ruleName) => ({
 
 export const enableRule = (payload) => {
   const rule = clonedeep(payload);
-  delete rule['queriedAt'];
-  delete rule['timestamp'];
 
   if (!rule.rule.value) {
     rule.rule.value = '';
@@ -804,8 +802,6 @@ export const enableRule = (payload) => {
 
 export const disableRule = (payload) => {
   const rule = clonedeep(payload);
-  delete rule['queriedAt'];
-  delete rule['timestamp'];
 
   if (!rule.rule.value) {
     rule.rule.value = '';
--- a/app/src/js/actions/index.js
+++ b/app/src/js/actions/index.js
@@ -744,7 +744,7 @@ export const getRule = (ruleName) => ({
     id: ruleName,
     type: types.RULE,
     method: 'GET',
-    path: `rules?name=${ruleName}`
+    path: `rules/${ruleName}`
   }
 });
 
--- a/app/src/js/reducers/providers.js
+++ b/app/src/js/reducers/providers.js
@@ -62,7 +62,7 @@ export default function reducer (state = initialState, action) {
   switch (action.type) {
     case PROVIDER:
       set(state, ['map', id, 'inflight'], false);
-      set(state, ['map', id, 'data'], assignDate(data));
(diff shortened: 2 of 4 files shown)


app/src/js/reducers/providers.js
```diff
@@ -62,7 +62,7 @@ export default function reducer (state = initialState, action) {
   switch (action.type) {
     case PROVIDER:
       set(state, ['map', id, 'inflight'], false);
-      set(state, ['map', id, 'data'], assignDate(data));
```
markdboyd: These lines were what was adding the unnecessary `queriedAt` property to the rule/provider objects and causing them to fail when sent as the body of a PUT requests to the API, since `queriedAt` is not included in the schema for either of those objects. As far as I can tell and based on my testing, this `queriedAt` property wasn't providing anything, so I just removed it

app/src/js/reducers/rules.js
```diff
@@ -59,9 +59,8 @@ export default function reducer (state = initialState, action) {
   const { data, id } = action;
   switch (action.type) {
     case RULE:
-      const foundRule = data.results.find((result) => result.name === id);
```
markdboyd: This was only necessary because the rules action was previously querying the GET endpoint listing all rules rather than using the GET endpoint for a specific rule. That is also what was causing there to be an extra `timestamp` property on the `rule` which also was causing PUT updates to fail.

app/src/js/actions/index.js
```diff
@@ -781,8 +781,6 @@ export const deleteRule = (ruleName) => ({
 
 export const enableRule = (payload) => {
   const rule = clonedeep(payload);
-  delete rule['queriedAt'];
-  delete rule['timestamp'];
```
markdboyd: Updating this action to get a rule using the individual rule GET endpoint means that an extraneous `timestamp` property is not returned, so we don't have to delete it

app/src/js/actions/index.js
```diff
@@ -781,8 +781,6 @@ export const deleteRule = (ruleName) => ({
 
 export const enableRule = (payload) => {
   const rule = clonedeep(payload);
-  delete rule['queriedAt'];
```
markdboyd: This extra property came from the reducer and seems unnecessary, so I prevented it from being added there and thus we won't have to remove it

see https://github.com/nasa/cumulus-dashboard/pull/611/files#diff-c33e901d446f106f94633d34f927341aL65

app/src/js/actions/index.js
```diff
@@ -781,8 +781,6 @@ export const deleteRule = (ruleName) => ({
...
-  delete rule['timestamp'];
 
   if (!rule.rule.value) {
```
markdboyd: This is still necessary I think because of a minor bug in our rules update endpoint. I can investigate that as part of this ticket

app/src/js/actions/index.js
laurenfrederick: Yup makes sense! Thanks for the explanation!

PR #614, opened 2019-12-11
Cumulus 1501 collection dropdown

jjmccoy:


app/src/js/components/DropDown/simple-dropdown.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 import React from 'react';
 import PropTypes from 'prop-types';
+// import './DropDown.scss';
```
kkelly51: Don't need this?

jjmccoy: Reason for comment out for now is that we will have wait on CSS module local imports for components. Currently, all CSS is global (main.scss) and in order to have local CSS styles for the components we need a CSS modifier -- we get a parse error if we try local. That's what I'm researching. We'll have to determine if we are going to replace Gulp/Browserify with Webpack or just add Webpack to work with Gulp/Browserify for static task and build management.

PR #618, opened 2019-12-23
Cumulus 1520

jjmccoy:
This is for CUMULUS 1520 and 1521 which is the Delete Collection Modal workflow. It contains the journey for a user deleting one collection and not batch or multiple collections.


app/src/index.js
```diff
@@ -19,6 +21,10 @@ class App extends React.Component {
       <div className='app'>
         { target !== 'cumulus' ? (
           <div className='app__target--container'>
+            { /*  Need to add alert function
```
kkelly51: What's happening here? Does this functionality need to go into this PR?

app/src/js/components/Button/Button.js
```diff
@@ -8,7 +8,8 @@ import {
   Tooltip
 } from 'react-bootstrap';
 
-import './Button.scss';
+// This causes a ParseError
+// import './Button.scss';
```
kkelly51: But do we still need this? And if we don't, can we completely get rid of it instead of just commenting it out?

jjmccoy: Yes, we still need it that's why it's commented out. This is the module issue with CSS and Gulp.

app/src/index.js
jjmccoy: Since I was working modals, I was looking at pop-ups as well and just was making notes as I went along so I don't forget to add that. It can be removed if you want for now. I have it saved locally.

app/src/js/components/DeleteCollection/DeleteCollectionModal.js
```diff
@@ -0,0 +1,76 @@
...
+    return (
+      <Modal
+        dialogClassName="delete-collection-modal"
+        show={this.props.show}
+        onHide={this.handleCancel}
+        centered
+        size="md"
+        aria-labelledby="modal__delete-collection-modal"
+      >
+        <Modal.Header className="delete-collection-modal__header" closeButton />
```
kkelly51: add the onCancel action to this close button

kkelly51: Currently I'm clicking the x and nothing happens

kkelly51: This is what I did with the bulk granules modal 

<Modal.Header className="bulk_granules-modal__header" closeButton onClick={this.cancel}></Modal.Header>

jjmccoy: It's working fine on this modal. It's the CollectionDeletedErrorModal and BatchAsyncCommand modals that are experiencing this issue. This should be able to bind with the handleCancel or handleClose function (whichever one we use). With react-bootstrap we should just use the closeButton and it knows when you declare the function before the render for both the "x" and the close/cancel button in the modal footer. I'll look at this.

PR #620, opened 2020-01-02
CUMULUS-1363: use npm instead of yarn

jennyhliu:


package.json
```diff
@@ -96,7 +97,7 @@
     "gulp-size": "^2.1.0",
     "gulp-sourcemaps": "^1.6.0",
     "gulp-uglify-es": "^1.0.4",
-    "gulp-useref": "3.1.5",
+    "gulp-useref": "^4.0.0",
```
markdboyd: I had some reservations about updating to this version, because it seems like `gulp-useref` is still relying on versions of `event-stream` that may be malicious:

https://github.com/jonkemp/gulp-useref/issues/256

Supposedly it's fine, but it still seemed disconcerting to me

jennyhliu: The latest `event-stream 4.0.1` no longer has malicious dependency `flatmap-stream` and should be safe, and `npm audit` doesn't report any errors.

markdboyd: OK, sounds good

.circleci/config.yml
```diff
@@ -6,38 +6,23 @@ jobs:
     steps:
       - checkout
```
flamingbear: Why did you remove the cache step?

flamingbear: I'm guessing because we went to use `npm ci` which nukes the node_modules directory before anything...

app/src/js/components/Executions/execution-status-graph.js
```diff
@@ -2,7 +2,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import dagre from 'dagre-d3v4';
+import dagre from 'dagre-d3';
 import * as d3 from 'd3';
 
 // dagre-d3v4 expects d3 to be attached to the window
```
flamingbear: We're not using dagre-d3v4 anymore.  Need to update comment.

.circleci/config.yml
flamingbear: Is this much slower than before?  It seems ok to me, just curious.

jennyhliu: > I'm guessing because we went to use `npm ci` which nukes the node_modules directory before anything...

Yes.  If the dependencies in `package-lock.json` don't match with `package.json`, `npm ci` will exit with error.
I guess it would be slower, but acceptable.

PR #622, opened 2020-01-07
Fix log search on granule detail page

chuckwondo:
Fix log search on granule detail page

On the granule detail page, when a user started typing in the
log search box, the UI broke and rendered a blank page due
to a React rendering "infinite loop."

CUMULUS-1679

### DIFF
--- a/app/src/js/components/Logs/viewer.js
+++ b/app/src/js/components/Logs/viewer.js
@@ -1,4 +1,6 @@
 'use strict';
+
+import truncate from 'lodash.truncate';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { interval, getLogs, clearLogs } from '../../actions';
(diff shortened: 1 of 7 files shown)


app/src/js/components/Logs/viewer.js
```diff
@@ -16,140 +18,160 @@ const noLogs = {
...
-  'Debug',
-  'Info',
-  'Warn',
-  'Error',
-  'Fatal',
-  'Trace'
-];
+// Use a Map to preserve order of entries.  Otherwise, when using a plain
+// object, during iteration over the keys (or entries), "numeric" keys are
+// yielded in numeric order, not in insertion order (Gah!).
```
jennyhliu: Can you point me the place where the keys were not ordered correctly before the changes?

app/src/js/components/Logs/viewer.js
```diff
@@ -16,140 +18,160 @@ const noLogs = {
...
     };
   }
 
-  componentDidMount () {
+  componentDidMount() {
     this.query();
   }
 
-  componentDidUpdate (prevProps) {
-    if (JSON.stringify(this.props.query) !== JSON.stringify(prevProps.query)) {
```
jennyhliu: This line is deleted.  It seems that this will create unnecessary queries and page updates.

app/src/js/components/Logs/viewer.js
jennyhliu: Ok.  I see the Log Level in granule page

chuckwondo: Right, this wasn't a problem before because there were no log level numbers before. This comment is specific to the new level numbers I added, which correspond to the change I made in the dropdown component to allow specifying option values that are different from the display values. In this case, this is because the API call wants numbers for the log levels, not names.

app/src/js/components/Logs/viewer.js
chuckwondo: No, because I also removed the call to `this.query`, which is what this condition was guarding.

PR #623, opened 2020-01-09
Cumulus-1463 Add Datepicker component

jennyhliu:


cypress/integration/main_page_spec.js
```diff
@@ -91,6 +91,31 @@ describe('Dashboard Home Page', () => {
       });
     });
 
+    xit('displays a date picker in metrics section', () => {
```
kkelly51: Does this not run because the xit? 
And you turned it off because it isn't currently on the home page?

jennyhliu: yes.  the test won't run with `xit`

PR #625, opened 2020-01-29
Cumulus 1102 add local express api app

flamingbear:
**NOTE**: ~There are still a few tests that appear to have difficulties, but this PR is ready for the beginning of reviews.  I will update the tests that are randomly failing.~ some strategic hard waits() seem to have solved this for now. 

**Summary:** Switch dashboard integration tests to use API express app instead of fake API

Addresses [CUMULUS-1102: Switch dashboard integration tests to use API express app instead of fake API](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1102)

## Changes

* Integration (cypress) tests and validation tests run in container against local Cumulus API.
* Adds ability to run dashboard against Cumulus localAPI.
    - Adds a number of docker-compose commands to be run via `npm run <command>`
        - `seed-database` - loads data fixtures into a running stack for testing
        - `start-localstack` - starts necessary backend for cumulus API. LocalStack + Elasticsearch
        - `stop-localstack` - stops same
        - `start-cumulusapi` - starts localstack and cumulus localAPI
        - `stop-cumulusapi` - stops same
        - `start-dashboard` - starts localstack, cumulus localAPI and dashboard
        - `stop-dashboard` - stops same
        - `e2e-tests` - starts starts localstack, cumulus localAPI, dashboard and cypress end to end tests.
        - `validation-tests` - starts starts localstack, cumulus localAPI, dashboard and validation tests.
        - `view-docker-logs`- helper to view logs for the currently running docker-compose stack.
* Removes fake-api.js.  The fake-api is removed in favor of running a Cumulus API locally.
* Fixes a number of bad cypress tests.

## PR Checklist

- [X] Update CHANGELOG
- [ ] Unit tests
- [X] Adhoc testing
- [X] Integration tests


cypress/plugins/seedEverything.js
```diff
@@ -0,0 +1,63 @@
...
+const { eraseDataStack } = require('@cumulus/api/bin/serve');
+
+const collections = require('../fixtures/seeds/collectionsFixture.json');
+const executions = require('../fixtures/seeds/executionsFixture.json');
+const granules = require('../fixtures/seeds/granulesFixture.json');
+const providers = require('../fixtures/seeds/providersFixture.json');
+const rules = require('../fixtures/seeds/rulesFixture.json');
+
+// Test values
+const stackName = 'localrun';
```
markdboyd: So I don't really like the Cypress code having to be aware of the internals of the local API, including these values or passing around the returned `esClient` and `esIndex`.

It seems possible to me that that we can expose a `resetDataStack` helper that handles fully resetting things without having to pass in any parameters.

I assume that the retrieval of `esClient` and `esIndex` is based on needing to index data after adding it to the DB. But I'm pretty sure we could also refactor those so that passing them in from here is unnecessary

markdboyd: I'll make some comments on the API PR

flamingbear: Actually I was trying to decouple it some so that this could be used by any person / client who wanted to, not just the tests with these values.  But I can clean it up some.

flamingbear: meaning you could use the localAPI with any bucket/index, not just the default ones

markdboyd: But the bucket is hardcoded and the index comes back from the helper which is based on the stackName? Is the intent here to be able to use these utilities with your real AWS stack and not just the local API?

flamingbear: It was mostly to be flexible with the names.  No intention of running with real AWS.  
But maybe that was overthinking.

cypress/integration/collections_spec.js
```diff
@@ -193,13 +190,15 @@ describe('Dashboard Collections Page', () => {
       cy.url().should('include', 'collections');
       cy.contains('.heading--xlarge', 'Collections');
       // verify the collection is now gone
+      cy.get('[data-value="http_testcollection___001"] > .table__main-asset > a').should('exist');
```
flamingbear: This is still giving me problems, and I will continue to look at it, but it's a race condition somewhere that I need to find.  the tests pass most of the time (and always when I do it by hand).

cypress/integration/collections_spec.js
```diff
@@ -208,10 +207,13 @@ describe('Dashboard Collections Page', () => {
       cy.contains('.modal-content .button__contents', 'Delete Collection')
         .should('be.visible').click();
 
+      cy.wait('@deleteCollection');
+
       // modal error should be displayed indicating that deletion failed
-      cy.get('.modal-content .error__report').should('be.visible');
+      cy.get('.modal-content .error__report', {timeout: 25000}).should('be.visible');
```
flamingbear: I don't actually believe this timeout does anything.. cargocult maybe?

cypress/integration/reconciliation_spec.js
```diff
@@ -9,18 +9,15 @@ describe('Dashboard Reconciliation Reports Page', () => {
   });
```
flamingbear: I added real reconciliation reports to the cumulus api.

.eslintrc
```diff
@@ -1,46 +1,48 @@
 {
   "extends": ["standard"],
-    "env": {
```
flamingbear: This "env" was indented wrong, most of this was is a result of that indentation.

cypress/integration/collections_spec.js
flamingbear: This problem is resolved by both forcing a refresh and increasing the timeout.

cypress/integration/collections_spec.js
flamingbear: Definitely not always working, will look into this this am.

README.md
```diff
@@ -105,15 +105,80 @@ To run the dashboard locally:
...
+```
+
+Once the docker app is running, If you would like some sample data (the same data that is used during integration tests), you can seed the database. This will load sample data into the application.
+```bash
+  $ npm run seed-database
+```
+
+You can run the dashboard locally (without docker):
+```bash
+  $ npm run serve
```
kkelly51: tell them to use APIROOT=http://localhost:5001

README.md
```diff
@@ -198,22 +263,38 @@ Then deploy the `dist` folder
...
-Run the test suite in another terminal:
+Serve the cumulus API (separate terminal)
+```bash
+  $ npm run serve-api
+```
 
+Serve the dashboard web application (another terminal)
 ```bash
-  $ npm run validation
+  $ npm run serve
```
kkelly51: jk here is where you should change the command to include the correct APIROOT

README.md
```diff
@@ -198,22 +263,38 @@ Then deploy the `dist` folder
...
+If you're just testing dashboard code, you can generally run all of the above commands as a single docker-compose stack.
+```bash
+  $ npm run start-dashboard
+```
+This brings up LocalStack, Elasticsearch, the Cumulus localAPI, and the dashboard.
+
+Run the test suite (yet another terminal window)
+```bash
+  $ npm run validate
   $ npm run cypress
```
kkelly51: And add the ES Host and LOCALHOST to the command, so the user doesn't have to deal with it

flamingbear: 8a0bde66

README.md
flamingbear: 8a0bde66

PR #626, opened 2020-02-04
fix collections modal btn stylings

jjmccoy:


app/src/js/components/DeleteCollection/BatchDeleteCollectionModal.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import Modal from 'react-bootstrap/Modal';
-import Button from '../Button/Button';
+// import Button from '../Button/Button';
```
flamingbear: Is this import for the next work (when Jenny's original PR gets worked on )? or should it just be deleted?

jjmccoy: Please keep. This import will work with my PR (Webpack - 1690). Gulp gives parse errors for the local import CSS that's in the Button.js component. It's best practice to import the button as a component so I have commented it out for now.

PR #628, opened 2020-02-06
Cumulus 1459 Small dashboard fix for asyncOps in ES

kkelly51:


app/src/js/components/Operations/overview.js
```diff
@@ -40,10 +40,10 @@ const tableHeader = [
 ];
 
 const statusOptions = {
-  Running: 'running',
-  Succeeded: 'succeeded',
-  Task_failed: 'task_failed',
-  Runner_failed: 'runner_failed'
+  Running: 'RUNNING',
```
laurenfrederick: Question about this and `typeOptions` below. When I look at the dropdowns in the dashboard I see `Task_failed` and `Runner_failed` which seems like it's coming from here. Is that supposed to get mapped to something and used here? Same for `typeOptions` I see the wrong strings.

kkelly51: What wrong strings are you seeing? 
I switched it from "running" to "RUNNING" All caps, because that is what goes to the API and matches it directly

app/src/js/components/Operations/overview.js
```diff
@@ -40,10 +40,10 @@ const tableHeader = [
...
+  Succeeded: 'SUCCEEDED',
+  Task_failed: 'TASK_FAILED',
+  Runner_failed: 'RUNNER_FAILED'
```
flamingbear: We should also change these two to be 'Task Failed' and 'Runner Failed' so that is what shows up in the dropdown instead of the underbarred.

kkelly51: Yup, I agree. Updated

app/src/js/components/Operations/overview.js
laurenfrederick: Looks good now!

flamingbear: Well Actually.... the status for failed might be wrong.
```
    {
      "output": "{\"name\":\"Error\",\"message\":\"Failed to fetch s3://mhs3-internal/600a34a6cec285d1490be3ecb1141f2300632147: The specified key does not exist.\",\"stack\":\"Error: Failed to fetch s3://mhs3-internal/600a34a6cec285d1490be3ecb1141f2300632147: The specified key does not exist.\\n    at fetchPayload (/home/task/index.js:45:11)\\n    at process._tickCallback (internal/process/next_tick.js:68:7)\"}",
      "createdAt": 1581097492955,
      "taskArn": "a787fbafc62331db8291d3d64b044774956160a7",
      "description": "Some description",
      "operationType": "ES Index",
      "id": "6ef86520-3f5b-4cf7-8dff-83818768a7e0",
      "status": "RUNNER_FAILED",
      "updatedAt": 1581097505287,
      "timestamp": 1581097509879
    },
```
I just got that from my testing.

flamingbear: Which probably means TASK_FAILED too.?

flamingbear: Look at me commenting on old code. Sorry. You already fixed it

PR #630, opened 2020-02-14
[CUMULUS-1549] Add Kibana link

bt6453:
* Replaces the `/logs` link with a link to Kibana
* If Metrics isn't set up, there is no link
* Remove now unused Logs overview page


app/src/js/components/Executions/execution-status.js
```diff
@@ -165,6 +169,11 @@ class ExecutionStatus extends React.Component {
           <dt>Logs:</dt>
           <dd><Link to={'/executions/execution/' + executionStatus.execution.name + '/logs'} title={executionStatus.execution.name + '/logs'}>View Execution Logs</Link></dd>
           <br />
+          { kibanaLink && kibanaLink.length ? (<div>
```
flamingbear: Can you get a verification from @jjmccoy that we want to show both links.  I think it's fine this way, but want a second opinion.

app/src/js/components/Header/header.js
```diff
@@ -53,10 +56,18 @@ class Header extends React.Component {
...
+    } else {
+      return <Link to={path[1]}>{path[0]}</Link>;
+    }
+  }
+
   render () {
     const { authenticated } = this.props.api;
-    const activePaths = paths.filter(pathObj => nav.exclude[pathObj[1].replace('/', '')] !== true);
-
+    const activePaths = paths.filter(pathObj => nav.exclude[pathObj[0]] !== true);
```
flamingbear: Is it possible to rename pathObj to something else, because those are arrays (yeah, I know they're technically Objects, but still that's misleading).

Also on this line https://github.com/nasa/cumulus-dashboard/pull/630/files#diff-2403c410025721c7ef089f0d869a3ea1R21. Can you change `/logs` to null? or something to show that it's not being used anymore?

app/src/js/utils/kibana.js
```diff
@@ -71,3 +71,17 @@ export const kibanaGatewayAccessSuccessesLink = (cumulusInstanceMeta) => {
...
+  const stackName = cumulusInstanceMeta.stackName;
+  return `${kibanaRoot}/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-48h,to:now))&_a=(columns:!(_source),index:${stackName},interval:auto,query:(language:kuery,query:''),sort:!('@timestamp',desc))`;
+};
+
+export const kibanaExecutionLink = (cumulusInstanceMeta, executionName) => {
+  if (!cumulusInstanceMeta || !cumulusInstanceMeta.stackName) return '';
+  if (!kibanaRoot) return '';
+  const stackName = cumulusInstanceMeta.stackName;
+  return `${kibanaRoot}/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-10y,to:now))&_a=(columns:!(_source),index:${stackName},interval:auto,query:(language:kuery,query:'executions:${executionName}'),sort:!('@timestamp',desc))`;
+};
```
flamingbear: It's not a problem for this PR, but it's probably time to look at https://www.npmjs.com/package/rison-node as a way to build our queries.  If I get a chance I might look at this in time for the Arch meeting.

app/src/js/components/Executions/execution-status.js
jjmccoy: I don't have an issue with redundancy but if you are going to 'View Execution Logs' and that takes you to executions/execution (little confused on where we are going from the 'Execution Details' if we are not going to 'Kibana' at this link) once and click on a log there doesn't that redirect you to Kibana? I think the user is going to wonder what's the difference between 'Logs' and 'Kibana Logs'.

flamingbear: so executions/execution is the execution detail for the given execution. (nice sentence).
Looks like this:

Screen Shot 2020-02-19 at 7 46 27 AM

flamingbear: It's just a view of the same execution logs inside of the dashboard.  I think it's harder to understand, but operators are going to have to get used to that view if we're giving them kibana.
Screen Shot 2020-02-19 at 7 48 25 AM

bt6453: After chewing it over I think one link is better (kibana if it's set up, dashboard page if it's not). Would they even use the regular dashboard page if Kibana is available?

jjmccoy: Okay cool. So I'm assuming that the 'Execution Logs' is more of a highlight with brief details for quick overview and the 'Kibana Logs' will be for more discovery?

app/src/js/components/Header/header.js
bt6453: Changed `pathObj`

So the `/logs` is used to determine a class name in the `className` method. It breaks the page if it's set to null. I changed it to `logs` which still works but maybe indicates it's no longer a link?

app/src/js/components/Executions/execution-status.js
flamingbear: @jjmccoy Yeah, I would think the link to kibana will take you to the records for the execution, but then you're right there ready to do discovery.

@bt6453 If kibana isn't available, what happens to that executions/execution link?  is there any data to show?

app/src/js/utils/kibana.js
```diff
@@ -71,3 +71,17 @@ export const kibanaGatewayAccessSuccessesLink = (cumulusInstanceMeta) => {
...
 };
+
+export const kibanaAllLogsLink = (cumulusInstanceMeta) => {
+  if (!cumulusInstanceMeta || !cumulusInstanceMeta.stackName) return '';
+  if (!kibanaRoot) return '';
+  const stackName = cumulusInstanceMeta.stackName;
+  return `${kibanaRoot}/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-48h,to:now))&_a=(columns:!(_source),index:${stackName},interval:auto,query:(language:kuery,query:''),sort:!('@timestamp',desc))`;
```
flamingbear:
```suggestion
  return `${kibanaRoot}/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-48h,to:now))&_a=(columns:!(_source),index:${stackName},interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc))`;
```

app/src/js/utils/kibana.js
```diff
@@ -71,3 +71,17 @@ export const kibanaGatewayAccessSuccessesLink = (cumulusInstanceMeta) => {
...
+};
+
+export const kibanaExecutionLink = (cumulusInstanceMeta, executionName) => {
+  if (!cumulusInstanceMeta || !cumulusInstanceMeta.stackName) return '';
+  if (!kibanaRoot) return '';
+  const stackName = cumulusInstanceMeta.stackName;
+  return `${kibanaRoot}/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-10y,to:now))&_a=(columns:!(_source),index:${stackName},interval:auto,query:(language:kuery,query:'executions:${executionName}'),sort:!('@timestamp',desc))`;
```
flamingbear:
```suggestion
  return `${kibanaRoot}/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-10y,to:now))&_a=(columns:!(_source),index:${stackName},interval:auto,query:(language:lucene,query:'executions:${executionName}'),sort:!('@timestamp',desc))`;
```

app/src/js/utils/kibana.js
flamingbear: Added to arch meeting.

app/src/js/components/Executions/execution-status.js
flamingbear: I tested for myself, it makes the request to the cumlusAPI/logs  endpoint and gets the same data back. Noice.

PR #624, opened 2020-01-10
CUMULUS-1463: Add Datepicker to Dashboard Home page

jennyhliu:

PR #621, opened 2020-01-03
Cumulus 1690

jjmccoy:
WIP Cumulus 1690 - Gulp to Webpack. Need assistance with errors regarding the webpack configuration.

PR #632, opened 2020-02-15
Updates Datepicker to react to state and update the state.

flamingbear:
Connects the Datepicker component to the redux store.

Dispatches actions based on user interaction with the Datepicker

Reduces actions to update store to reflect the state of the Datepicker UI.

Adds enough state to router to create shareable urls.

Things to finish ticket:

- Update cumulus API calls to examine the datepicker state before fetching data and append correct information for filtering based on user desired times.
- Trigger those updates for pages that need them.
- Update router to keep the URL state on each page.  Right now it only displays on the home page where the datepicker lives.
  - state does persist so that if you navigate away, and back the Datepicker retains it's state.


flamingbear:
after this is approved and merged #624 should be closed in favor of this one.

app/src/js/components/Datepicker/Datepicker.js
```diff
@@ -1,103 +1,94 @@
...
+  updateRouterWithNewProps (newProps) {
+    const urlDateFormat = 'YYYYMMDDHHmmSS';
+    const urlProps = ['endDateTime', 'startDateTime'];
+
+    urlProps.map((time) => {
+      let urlValue = '';
+      if (newProps[time] !== null) {
+        urlValue = moment.utc(newProps[time]).format(urlDateFormat);
+      }
+      updateRouterLocation(this.props.router, this.props.location, time, urlValue);
```
dopeters: With the new webpack stuff, `udpateRouterLocation` will no longer work. Instead, we are using the higher order component `withQueryParams` (https://github.com/woltapp/react-router-query-params). We should keep that mind if this gets merged before https://github.com/nasa/cumulus-dashboard/pull/621/

app/src/js/components/Datepicker/Datepicker.js
```diff
@@ -173,32 +163,32 @@ export class Datepicker extends React.PureComponent {
...
+              </li>
+              <li data-cy='endDateTime'>
+                { this.renderDateTimeRange('endDateTime') }
+              </li>
+              <li className="selector__hrformat" data-cy='hourFormat'>
+                { this.renderHourFormatSelect() }
+              </li>
+            </ul>
+          </div>
+          <div className="datetime__wrapper">
```
dopeters: I know this didn't arise from this PR, but we might want to consider putting all of the `.datetime` contents into `.datetime__wrapper` and adjusting the styles accordingly so that we can avoid some of the absolute/relative positioning that's going on with this component.

app/src/js/components/Datepicker/Datepicker.js
```diff
@@ -173,32 +163,32 @@ export class Datepicker extends React.PureComponent {
...
+                { this.renderDateRangeDropDown() }
+              </li>
+              <li data-cy='startDateTime'>
+                { this.renderDateTimeRange('startDateTime') }
+              </li>
+              <li>
+                <span> to </span>
+              </li>
+              <li data-cy='endDateTime'>
+                { this.renderDateTimeRange('endDateTime') }
```
dopeters: Is there a plan to add validation if the end date is before the start date?

app/src/js/components/Datepicker/Datepicker.js
jjmccoy: We needed the absolute/relative positioning for the calendar and clock overlays. Since we will be hiding them they will no longer be an issue. We can refactor the style but we can put that in another ticket.

app/src/js/components/Datepicker/Datepicker.js
flamingbear: I wasn't going to, and the thought occurred to me, we could handle it in a number of ways.  I thought I'd see what people liked and didn't like about the functionality. If we need that for this PR, I can add it.

app/src/js/components/Datepicker/Datepicker.js
flamingbear: I'm calling this, future-Matt's problem.

app/src/js/components/Datepicker/Datepicker.js
flamingbear: Thanks, this is coming after, I'll check on that.

app/src/js/components/home.js
```diff
@@ -43,14 +46,33 @@ import {
...
+  if (!isEmpty(queryParams)) {
+    for (const value in values) {
+      if (urlDateProps.includes(value)) {
+        values[value] = moment.utc(values[value], urlDateFormat).toDate();
+      }
+    }
+    values.dateRange = {value: 'Custom', label: 'Custom'};
+    props.dispatch({type: 'DATEPICKER_DATECHANGE', data: {...props.datepicker, ...values}});
+  }
+};
```
dopeters: Nothing actionable, but we may want to eventually move this outside of the home component if we want to be able to use a shareable date on other pages.

flamingbear: Oh yeah good catch! I remembered this at one point and never asked anyone about it.  It does seem like we could move it to the Datepicker.js file.  I will do that in the next PR if it's easy/possible.

app/src/css/vendor/datetimepicker/_datetimepicker.scss
```diff
@@ -0,0 +1,13 @@
...
+******************************************************************/
+/*
+All React Date Time Picker imports
+*/
+
+/* Custom Overrides */
+@import "../datetimepicker/dtpoverrides";
+
+/* Default Styles */
+// @import "../datetimepicker/dtpstyles";
```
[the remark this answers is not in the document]
flamingbear: @jjmccoy ?  (I can remove it in the follow on also if need be)

flamingbear:
> This looks good to me! Nothing actionable here. Though just want to note that there's a little bit of funkiness going on with the styling of the box when the inputs are empty v when they have a value.
> Notice how the whitespace around the input changes. Not a huge deal, but if we do restyle the calendar, we might want to keep that in mind to change then.
> 
> Screen Shot 2020-02-21 at 12 46 07 PM
> 
> Screen Shot 2020-02-21 at 12 46 13 PM

Do you think that's just a fontsize difference between the placeholders and the real values? because it seems like the same number of characters?

app/src/css/vendor/datetimepicker/_datetimepicker.scss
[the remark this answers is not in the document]
jjmccoy: You can remove the comment that `@import` is not needed.

The meridian is the issue above input. With "--" and it has a default space afterwards for some reason. I believe it was supposed to be an option dropdown icon there but it's not showing and I don't know why removes the space (which should be an icon) when "AM" or "PM" is selected.

app/src/js/components/home.js
flamingbear: I just added this to my next PR that also has the speed improvements.

PR #633, opened 2020-02-17
Cumulus 1729 speed up testing

flamingbear:
Updates Cypress Tests to make them faster.

There is one functional changes to this PR.

Now we are only reseting the localstack state when a test has
modified it.  Before we did it before every test, this was taking about 6 seconds each time and added up to a lot of useless cycles.  This change saves about 4min per run.

The second change has been reverted.

~The second change is that jwt tokens are now created via the code in @cumulus/api rather than in copy-pasted code that wasn't kept up to date. As a result, in order to create tokens that are valid on the api, we need to share the same secret between the API and the dashboard.  It's hard coded to "myDashboardSecret"  (I'm open to ways of making this private or more configurable, but doesn't seem very important. It might be nicer to make it DRY-er).~

[  ] before merging, this should have it's target branch changed to develop


flamingbear:
replaced by #640

PR #635, opened 2020-02-20
Cumulus 1502

dopeters:
This adds a copy button to an individual collection page.
Adjusted some css on the buttons to display it properly.
Also, adding in redux-logger for easier keeping track of the redux store.

To test:
1. Go to a collection
2. Click copy
3. Verify you are taken to /collection/add with the info from the collection
4. Change the version or name
5. Submit
6. Verify new collection is made


CHANGELOG.md
```diff
@@ -29,15 +26,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - **CUMULUS-1463**
   - Add Datepicker Component
 
+- **CUMULUS-1502**
+  - Add copy collection button
+
+- **CUMULUS-1534**
+  - Add confirmation modal when adding a new rule
+
```
chuckwondo: Excellent. Thanks for ordering these!

app/src/js/components/AddRaw/add-raw.js
```diff
@@ -37,6 +37,16 @@ const AddRaw = ({
...
+  useEffect(() => {
+    setRecord(record => {
+      const data = JSON.stringify(defaultValue, null, 2);
+      return {
+        ...record,
+        data
+      };
+    });
+  }, [defaultValue]);
+
```
chuckwondo: I know that the call to `useEffect` on line 50 is not part of your commits, but when I run `npm run lint`, I get the following warning. Perhaps you can implement the suggested fix to eliminate the warning.
```
/Users/chuck/src/nasa/cumulus-dashboard/app/src/js/components/AddRaw/add-raw.js
  50:3  warning  React Hook useEffect contains a call to 'setRecord'. Without a list of dependencies, this can lead to an infinite chain of updates. To fix this, pass [pk, state.created, error, getBaseRoute, history, record] as a second argument to the useEffect Hook  react-hooks/exhaustive-deps
```

app/src/js/components/Collections/add.js
```diff
@@ -1,40 +1,67 @@
...
-  }
-}
+  const getBaseRoute = (pk) => {
+    let id = pk || collectionId;
+    if (id && id !== 'unknown') {
+      return collectionHref(id);
+    } else {
+      return '/collections';
+    }
+  };
```
chuckwondo: Ignorable suggestion, but here's a way to simplify the function above:
```
const getBaseRoute = (pk = collectionId) =>
  (pk && pk !== 'unknown') ? collectionHref(pk) : '/collections';
```

cypress/integration/collections_spec.js
```diff
@@ -138,6 +138,38 @@ describe('Dashboard Collections Page', () => {
...
+      cy.visit(`/collections/collection/${name}/${version}`);
+      cy.contains('a', 'Copy').as('copyCollection');
+      cy.get('@copyCollection')
+        .should('have.attr', 'href')
+        .and('include', '/collections/add');
+      cy.get('@copyCollection').click();
+
+      cy.contains('.heading--large', 'Add a collection');
+
+      cy.wait(500);
```
chuckwondo: Can we eliminate this hard-coded wait time? What is happening that is requiring a wait here? Are we waiting on the collection to be fetched? If so, can we replace this with `cy.wait('@getCollection')`?

app/src/js/components/Collections/add.js
dopeters: Good call, this is so much cleaner! Changing.

app/src/js/components/AddRaw/add-raw.js
dopeters: Done

cypress/integration/collections_spec.js
dopeters: This one was tricky because it wasn't waiting for a request, but a state update. he `defaultValue` prop changes once we get the schema, and it wasn't seeing the change without the `wait` Since `.contains()` will retry until it finds the element in the DOM, I was able to work around this by instead checking that `.ace__variable` containing the value "name" existed in the DOM.

app/src/js/components/AddRaw/add-raw.js
chuckwondo: What happened to the list of deps?

cypress/integration/collections_spec.js
chuckwondo: Perfect!

PR #639, opened 2020-02-22
Cumulus 1729 filter api calls with correct date information

flamingbear:
Filters Cumulus API calls when state.datepicker has startDateTime or endDateTime values.

These list action creators are updated with the current state:
    `listCollections`
    `listExecutions`
    `listGranules`
    `getLogs`
    `listOperations`
    `listPdrs`
    `listProviders`
    `listRules`

As a result when these are called, the Cumulus API returns a filtered set of results and only the these are displayed on the dashboard.


app/src/js/utils/datepicker.js
```diff
@@ -1,3 +1,28 @@
...
+  { dateProp: 'endDateTime', filterProp: '__to' }
+];
+
+export const urlDateProps = matchObjects.map((o) => o.dateProp);
+
+/**
+ * Build a timefilter object that will be passed to the cumulus core API calls as part of the querystring.
+ *
+ * @param {Object} datepicker - redux datepicker state.
+ * @return {Objecte} object suitable to passing to a request querystring that will contain
```
bt6453: Object misspelled here as `Objecte`

app/src/js/actions/index.js
```diff
@@ -542,14 +545,17 @@ export const getCount = (options) => ({
...
-    type: types.PDRS,
-    method: 'GET',
-    url: url.resolve(root, 'pdrs'),
-    qs: Object.assign({ limit: pageLimit }, options)
-  }
-});
+export const listPdrs = (options) => {
+  return (dispatch, getState) => {
+    const timeFilters = fetchCurrentTimeFilters(getState().datepicker);
+    dispatch({[CALL_API]: {
```
markdboyd: I could be wrong, but I think you want to `return dispatch`, otherwise you wouldn't be able to do `.then` on `listPdrs` or any of these action creators if you wanted to later

flamingbear:
The errors on the home page aren't updating, and that will be fixed in a separate PR.

app/src/js/actions/index.js
flamingbear: You're right.  a52de30f21f028c18a790305912b8af3177cd036

PR #644, opened 2020-02-26
Give each MMT link a waitable route.

flamingbear:
Wait for those routes when testing MMT links.

Removes unused CMR fixture for TESTCOLLECTION.


flamingbear:
> Nice!

We will see.  It's a start and it's better than it was.

PR #646, opened 2020-02-26
Save Cypress videos

flamingbear:
Saves cypress videos as artifacts in circleCI.

Moves the wait for MMT, (but that doesn't actually seem to be working on my other branch.)

PR #648, opened 2020-02-27
Cumulus 1729 persist url query parameters

flamingbear:
### keep any query parameters when you navigate to different pages. 

Persists query parameters in the URL when navigating between pages.

- changes all imports from react-router to imports from react-router-dom.

- changes the value in setQueryParams from "" to undefined when clearning the value, this removes the query param from the url rather than setting it to nothing.
  so `?status=running&startDatetime=201001021500` => `?startDatetime=201001021500`
  rather than `?status=&startDatetime=201001021500`

- also includes creating videos in circleCI. in case this gets merged before #646 

 - It will also keep selector query params between pages. Like selecting “completed” from the status dropdown on granules and then navigating to executions, the status field will have completed selected and the results will be filtered


flamingbear:
chat from #cumulus-ox-dev
Danielle Peters  2:50 PM
don't kill me @Matt Savoie but.....looking at the persisting query params PR and am thinking...do we actually want/need to persist them? changing the the date with the datepicker updates the state so we should be able to just grab the dates from there. and i don't love keeping the non date params (i.e. status, etc) across different pages. if we do want the date ones to persist, i think we should filter out the other ones

Matt Savoie  2:52 PM
The issue would be if you wanted to show another user a filtered list on the granules page, you need to have all of the params in the URL if you want to build the state up from the initial url.  So they wouldn’t be shareable.
2:52
I can be convinced to filter off the non-dates ones pretty easily though.  and I don’t have super strong opinions on the dates

Danielle Peters  2:54 PM
hmm i gotcha. we could do something where we first check the state and if it's empty, we then grab the query params or vice versa

Matt Savoie  2:55 PM
? Not sure what you mean.
2:55
I think on initialization of a component it should grab what’s in the URL and never think about it again, except to keep it updated

Danielle Peters  2:58 PM
yeah i agree. i'm just mean that when it initializes, we'll either have a value in the state or not. if we already do, that means that we've been in the app and have done things, so we can use whatever's there in the same way we're using the query params without having to explicitly have params (edited) 

Matt Savoie  2:58 PM
gotcha.
3:00
I guess we still need to know if we need/want shareable URLs.  I thought that was a requirement, but ?.

Danielle Peters  3:00 PM
ohhhh i see what you're saying about shareable

Matt Savoie  3:00 PM
If we’re going to handle start/end dates differently from the other fields, we should probably have a convo on what behavior we want.

Danielle Peters  3:00 PM
like we need to generate the url itself (edited) 

Matt Savoie  3:01 PM
More like, I have this problem here.  /check it out.   http://localhost:3000/executions?endDateTime=20100227215762&startDateTime=2100128215762
3:01
and when you load the app it should use those values to initialize itself.
3:01
As a benefit, you can actually tweak the url directly and get the updates to the components.
3:02
so you can put your own start/end date into the url and the datepicker would pick up those changes.
3:02
Right now the qp’s are persisted, but don’t update the state except on teh home page I don’t think.
3:03
but that would change when we get datepickers on all the pages that use them.

Danielle Peters  3:03 PM
i gotcha. ok that all makes sense. though, shouldn't we be able to use the params for api calls even if the datepicker isn't on the page?

Matt Savoie  3:04 PM
Yes, that’s happening now, but not with a sharable URL because the state isn’t updated (like if you used that link above)


4 replies
Last reply today at 3:07 PMView thread

Danielle Peters  3:04 PM
also, since i can see why we do want to keep these params, we're gonna need to handle the sidebar links as well as the links within tables, right?

Matt Savoie  3:04 PM
Yup, that’s true
3:04
skipped them completely

Danielle Peters  3:05 PM
cool. alright thanks for talking that out, it's a lot clearer to me now
3:06
do you want to address the sidebar and other links in this PR or a new one? it's good to go besides that

Matt Savoie  3:08 PM
So I can probably find the sidebar links, which others are you thinking?

Danielle Peters  3:09 PM
if for example you navigate to /collections, and then click on one in the table, the query params get lost. might be annoying to deal with all of those

Matt Savoie  3:10 PM
lemme see

Danielle Peters  3:13 PM
hmm i wonder if for those cases we could do something like check the state for dates and append them to the url if they exist. idk if that's the best solution though

Matt Savoie  3:14 PM
Well, I can see arguments forleaving them off, because it’s easy, and the state is still there so when you go to another page, it would be able to use the state to update datepickers and the URL again?
3:15
so maybe the best thing is to have pages that show a datepicker ensure the URL matches the state. and leave those alone?
3:17
right now, the URL loses the params, but when you navigate to the homepage, the datepicker is still set properly.  I guess I need to think on that

Danielle Peters  3:19 PM
yeah i mean, you'd know more than I would when/where the params are useful, so that's totally fine by me :shrug:. but yeah on that last point, sounds like we should at least sync the state and url on the pages w/ datepickers

Matt Savoie  3:22 PM
i feel pretty strongly that the URL should reflect the state, and we should only be pulling information off of it when we have to. (on first load).  But yes, maybe there’s something I can do to keep those queryParams in the URL on pages where the rendering is filted by them.
3:22
Ok, thanks for talking that out.

Danielle Peters  3:23 PM
yeah absolutely, totally agreed

Matt Savoie  3:24 PM
Is that a this ticket or another I guess also.

Danielle Peters  3:27 PM
up to you!

Matt Savoie  3:27 PM
cumulus-1466 and 1468 are already working . but 1468 is specific to updating the granules page.
3:27
Hmm.  Ok. well its late out there, you should go home. :slightly_smiling_face:
3:27
I’ll be back tomorrow afternoon.
new messages

Danielle Peters  3:32 PM
haha alright, well i can approve the current PR. though i do still think we should strip the non date params from a ux stand point (i probably don't want to see the search query i made for collections on the providers page for instance)

flamingbear:
Made a follow up ticket to get issues left after this.  https://bugs.earthdata.nasa.gov/browse/CUMULUS-1773

PR #663, opened 2020-03-05
Cumulus 1504

jjmccoy:


cypress/integration/granules_spec.js
```diff
@@ -28,11 +28,11 @@ describe('Dashboard Granules Page', () => {
...
       cy.get('.overview-num__wrapper ul li')
         .first()
-        .contains('li', '10 Completed')
+        .contains('li', 'Completed')
         .next()
-        .contains('li', '3 Failed')
+        .contains('li', 'Failed')
         .next()
-        .contains('li', '3 Running');
+        .contains('li', 'Running');
```
flamingbear: Can we keep the tests for the numbers?  Those are still important aren't they? 
Screen Shot 2020-03-06 at 5 26 01 PM

flamingbear: @jjmccoy I peeked at this and I'm not sure they're actually important, but they were not too difficult to add back.   https://github.com/nasa/cumulus-dashboard/pull/676

PR #666, opened 2020-03-05
Add ability to render custom components in bulkActions

dopeters:
Proof of concept for "dummy" bulkAction buttons


jjmccoy:
It's a go! Working for all `bulkaction` buttons for Collections and Granules.

PR #673, opened 2020-03-06
Adding git dependency and SERVED_BY_CUMULUS_API flag to build with do…

amarouane-ABDELHAK:
…cker


flamingbear:
Hi I'm going to pull this into another branch so that the CI tests can run.  I think you should still get the commit credit. Thanks!

PR #675, opened 2020-03-07
Update api-version's reducer to not modify state.

flamingbear:
This PR came about after talking with Mark about reducers and mutating state.

This particluar PR is an example of what I was thinking of doing for all of the reducers.

The command `state = {...state}` creates a new shallow object of the state, and then returned it at the end.

React-redux does a shallowEqual to determine if an object has changed, this will prevent a rerender if it is not changed.  For this reducer, there are no functional differences.  The state passed in to the reducer is completely flat, and shallowEqual used by react-redux (https://github.com/reduxjs/react-redux/blob/master/src/utils/shallowEqual.js) will return unchanged even though the returned object is a strictly different object.  statein !=== stateout.

But. That said, we have a number of places where we are mutating state in reducers, which means that the shallowCompare is returning equal when the state is different and the data should be updated, but is not.  This is the case for every time we call `set(object, [multiple keys], value)`

I was hoping that we could agree to a standard reducer pattern that would prevent both over-rendering (returning objects that are different even when no changes have occurred) but also under-rendering (and BUGGY behavior) when state is mutated in a reducer.

I have another branch that fixes a lot of these problems, but wanted to get eyes on this particular solution before moving forward.

Mark B suggested he liked returning a new object from each case statement better than setting `let newState=null` and then returning `return newState || state` at the end.  for example.

### DIFF
--- a/app/src/js/reducers/api-version.js
+++ b/app/src/js/reducers/api-version.js
@@ -12,24 +12,36 @@ export const initialState = {
 };
 
 export default function reducer (state = initialState, action) {
-  state = {...state};
+  let newState = null;
   switch (action.type) {
     case API_VERSION:
-      state.versionNumber = action.payload.versionNumber;
-      state.warning = '';
+      newState = {
+        ...state,
+        versionNumber: action.payload.versionNumber,
+        warning: ''
+      };
       break;
     case API_VERSION_ERROR:
-      state.apiVersion = action.payload.error.message;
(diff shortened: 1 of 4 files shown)


jjmccoy:
LGTG Screen Shot 2020-03-10 at 1 33 32 PM

flamingbear:
consensus was that this was not it.  They liked returning copies from inside the case statements.

PR #685, opened 2020-03-12
[CUMULUS-1693] [CUMULUS-1694] Bulk delete collections and BatchAsyncCommands updates

dopeters:
This updates the batch delete collections workflow.

Now, if a user tries to delete a collection that has granules, the modal will update with information indicating which collection contains granules and the confirm button will redirect to the granules page.

Changes to Batch Async Commands: changes modal to use the DefaultModal component. Then, added a modalOptions prop which can be set up in the table config containing the batch actions. This is an object containing any props to be passed to the modal. 

Also noticed an issue where errors weren't being handled correctly because they weren't being passed to `onBulkActionError`. Also fixed a similar error where `bulkActionError` was being cleared on success even if there should be an error displayed, so I fixed that by passing in the error to the success function as well to determine whether or not to clear it.

Finally, found an issue with the selection state management due to the new table implementation. I was able to fix by using a `clearSelection` prop that when changed (i.e. after a bulk action success or failure), triggers an effect which toggles all checks off.

-----
This should also include changes necessary for 1694, which is a refactoring of BatchAsyncCommands. To add a simple button, in the table-config file where the bulkActions array is being created, you can include an item with a `Component` property. If there is a value for that property, it will be used instead of the BatchAsyncComponent. This allows for Add buttons that redirect to the add collections/rule page, or other such buttons that need to do a simple action.

Additionally, with the `getModalOptions` function passed to BatchCommands, you can dynamically change the modal content depending on your needs.


jjmccoy:
Bulk delete collections workflow looks good just checking the overall batch async commands for other components.

PR #689, opened 2020-03-16
Cumulus-1526 Copy a Rule

kkelly51:


app/src/js/components/Rules/add.js
```diff
@@ -45,11 +30,48 @@ ModalBody.propTypes = {
...
+      name: '',
+      version: ''
+    },
+    meta: {},
+    rule: {
+      type: '',
+      value: ''
+    },
+    state: 'ENABLED'
+  });
```
dopeters: Might be a little easier to read if we kept the filler object outside of the component with a different variable name and just set it here like `const [defaultValue, setDefaultValue] = useState(defaultRuleObject);`, but doesn't really matter

app/src/js/components/Rules/add.js
```diff
@@ -45,11 +30,48 @@ ModalBody.propTypes = {
...
+  const { state: locationState } = location;
+  const { name } = locationState || {};
+  const { rule: ruleSchema } = schema || {};
+  const { map: rulesMap } = rules || {};
+  const isCopy = !!name;
+  let title;
+  (isCopy) ? title = 'Copy a rule' : title = 'Add a rule';
```
dopeters: We could make this more concise with `const title = isCopy ? 'Copy a rule' : 'Add a rule'`

We don't have this for the copy collection but perhaps we should add it? 

Also, I'm wondering if it's worth changing the copy to something like \`Copying rule ${name}\` just to further orient the user, but I don't have a strong opinion either way 🤷‍♀

cypress/integration/rules_spec.js
```diff
@@ -111,6 +111,40 @@ describe('Rules page', () => {
...
+      cy.contains('.table .tbody .tr a', testRuleName)
+        .and('have.attr', 'href', `/rules/rule/${testRuleName}`)
+        .click();
+
+      cy.contains('.heading--large', testRuleName);
+      cy.contains('.button--small', 'Copy').click();
+      cy.contains('.heading--large', 'Copy a rule');
+
+      const newName = 'testRule2';
+      cy.editJsonTextarea({ data: { name: newName }, update: true });
```
dopeters: This test is occasionally failing. What I've found is that sometimes the rerender with the default content has not happened before this editJsonTestArea call. Adding in a `contains` command for the fields that should be present seems to fix it because of cypress' built in retry functionality. 
```suggestion
      cy.contains('.ace_string', testRuleName);
      cy.editJsonTextarea({ data: { name: newName }, update: true });
```

app/src/js/components/Rules/add.js
kkelly51: I did think about that, add the "copying rule ${name}" But I was thinking once they get to this page, it's more about the new rule and We should encourage them to change the name anyway

app/src/js/components/Rules/add.js
kkelly51: sure

app/src/js/components/Rules/add.js
dopeters: Yeah agreed. I wonder if we need to have a title change at all then? Like, regardless, they're still "adding" a rule, so it might be good to indicate that. But again, I don't really have a strong opinion on it.

cypress/integration/rules_spec.js
```diff
@@ -122,8 +122,16 @@ describe('Rules page', () => {
       cy.contains('.heading--large', 'Copy a rule');
 
       const newName = 'testRule2';
+      cy.contains('.ace_variable', 'name');
```
dopeters: My thinking is that since 'name' will be present for the default empty rule object and the copied rule, and the copied rule will only be present after a rerender, we need to look for something specific to the copied rule.

PR #691, opened 2020-03-18
Cumulus 1790 use recent dropdown

flamingbear:
Adds a "Recent" dropdown to the datepicker options.

Functionality is changed so that the default is "recent" with also sets 24hour back as the start time.
All other selections will act as before set both start and end times.
selecting "recent" will leave the start time at whatever you had, and set end time to Now.

Let's talk about this.


app/src/js/components/Datepicker/Datepicker.js
```diff
@@ -169,18 +171,23 @@ class Datepicker extends React.PureComponent {
...
+            hourPlaceholder='HH'
+            locale={locale}
+            monthPlaceholder='MM'
+            minutePlaceholder='mm'
+            name={name}
+            onChange={(value) => this.handleDateTimeRangeChange(name, value)}
+            value={utcValue}
+            yearPlaceholder='YYYY'
+          />
+        ) : <div className="datetime__placeholder react-datetime-picker__wrapper react-datetime-picker__inputGroup">{placeholder}</div>
```
flamingbear: this is Matt having no idea how to css

flamingbear:
closing in favor of #693

PR #699, opened 2020-03-23
Cumulus 1499

jjmccoy:
Introducing the DateTimePicker search module in the header of non-home page components.


app/src/js/components/Operations/index.js
```diff
@@ -1,17 +1,33 @@
...
+            <ul className='datetimeheader'>
+              <li>
+                <div className='datetimeheader__content'>
+                  <h1 className='heading--xlarge'>Operations</h1>
+                </div>
+              </li>
+              <li>
+                <DatePickerHeader onChange={this.query} />
+              </li>
+            </ul>
```
dopeters: Seems like we should be able to make this its own component (or included as part of the `DatePickerHeaader` component) with the title passed as a prop so we don't have to keep rewriting the markup around the header

jjmccoy: - For accessibility it would be best to keep the the input labels.
- Calendar text: Good catch. I will look at the calendar inherit color stylings.
- That would mean we would have to support two different components when we update `DatePicker` then we would have to update` DatePickerHeader` too. For now, I'm okay with `DatePicker` passing to `DatePickerHeader`. I'll have @chuckwondo  look at your suggestion about markup.

dopeters: - Should we maybe add a label above the first dropdown then to get them aligned?
- Great!
- I'm thinking it would be best to just replace the current DatePickerHeader with something like this so we wouldn't have to make changes in multiple places:

```
const DatePickerHeader = ({ heading, onChange }) => {
  return (
      <ul className='datetimeheader'>
        <li>
         <div className='datetimeheader__content'>
          <h1 className='heading--xlarge'>{heading}</h1>
        </div>
      </li>
      <li>
        <Datepicker hideWrapper={true} onChange={onChange} />
      </li>
    </ul>
    );
}
```
Btw: changed it to a function component rather than a class component in that example because I prefer it, but I won't push too hard for that 😂

jjmccoy: - Yeah that would be good to add. 
- Maybe Chuck will grant it for you. 😂

chuckwondo: @dopeters Yes, please feel free to change to a function component! I'd prefer that too.

chuckwondo: @dopeters @jjmccoy I'd be happy with one of you changing the DatePickerHeader component to also include the additional elements and to remove them from the other components to make things DRY.

dopeters: Awesome, seeing the changes now, thanks @chuckwondo and @jjmccoy! Last quick thing - the days of the week are also showing up as white text on the non-homepage pages. Once that's fixed, I'll approve.

jjmccoy: Thanks. Just fixed the weekdays styling.

PR #703, opened 2020-03-25
Cumulus 1796 prevent state mutations in Dashboard reducers

flamingbear:
The pr adds @reduxjs/toolkit and uses it's createReducer helper function to replace the existing reducer functions with ones that don't mutate the input state.

- createstore is replaced with the toolkit's configureStore
  https://redux-toolkit.js.org/api/configureStore
  
- router middleware is updated to to include new defaults for non-production builds.
  https://redux-toolkit.js.org/api/getDefaultMiddleware#included-default-middleware

- Updated each reducer to use createReducer to ensure no state mutations.
  https://redux-toolkit.js.org/api/createReducer
  
- Adds a number of reducer tests.


cypress/integration/providers_spec.js
```diff
@@ -110,6 +110,7 @@ describe('Dashboard Providers Page', () => {
     });
 
     it('should edit a provider', () => {
+      // TODO [MHS, 2020-03-24]  UH. what's going on there. with submit then success loop.
```
flamingbear: I would like to fix this, but not in this PR.  Unless someone demands it.

cypress/integration/providers_spec.js
```diff
@@ -137,11 +138,12 @@ describe('Dashboard Providers Page', () => {
         .type(host);
 
       cy.get('form div button').contains('Submit').click();
```
flamingbear: the `submit` doesn't take you to overview now.  you have to click submit, then success.  And I haven't had a chance to see why.

app/src/js/reducers/stats.js
```diff
@@ -14,55 +16,60 @@ import {
...
-      break;
-    case COUNT_INFLIGHT:
-      count = { data: state.count.data, inflight: true, error: state.count.error };
-      nextState = Object.assign({}, state, { count });
-      break;
-    case COUNT_ERROR:
-      count = { data: state.count.data, inflight: false, error: action.error };
-      nextState = Object.assign({}, state, { count });
-      break;
+  [STATS]: (state, action) => {
```
flamingbear: add comment about why this is janky

flamingbear: 0d8c33f1

app/src/js/reducers/collections.js
```diff
@@ -48,105 +42,116 @@ export const initialState = {
...
-      set(state, ['map', id, 'inflight'], false);
-      set(state, ['map', id, 'error'], action.error);
-      break;
+export default createReducer(initialState, {
+  [COLLECTION]: (state, action) => {
+    const { id, data } = action;
+    const colName = id.split('___');
+    const collection = data.results.find(function (element) {
+      return element.name === colName[0];
+    });
```
chuckwondo: nit: While you're in here, perhaps you could import `deconstructCollectionId` from `api/lib/utils.js` and change these line to the following:

```
const { id, data } = action;
const { name } = deconstructCollectionId(id);
const collection = data.results.find((element) => element.name === name);
```

No biggie if you choose not to. Just thought it might be nice to remove the "hard-coded" separator.

app/src/js/reducers/datepicker.js
```diff
@@ -49,25 +50,26 @@ const recentData = () => {
...
+        return {...state, ...computeDateTimeDelta(data.dateRange.value), ...data};
+    }
+  },
+  [DATEPICKER_DATECHANGE]: (state, action) => {
+    const { data } = action;
+    return { ...state, ...data };
+  },
+  [DATEPICKER_HOUR_FORMAT]: (state, action) => {
+    const { data } = action;
+    return { ...state, ...{hourFormat: data} };
```
chuckwondo: nit: no need for spreading a literal element; just do this:
```
return { ...state, hourFormat: data }
```

app/src/js/reducers/granule-csv.js
```diff
@@ -5,30 +5,22 @@ import {
...
-      break;
+export default createReducer(initialState, {
+  [GRANULE_CSV]: (state, action) => {
+    return {...state, ...{ data: action.data, inflight: false, error: null }};
+  },
+  [GRANULE_CSV_INFLIGHT]: (state) => {
+    return {...state, ...{ inflight: true, error: state.error }};
+  },
+  [GRANULE_CSV_ERROR]: (state, action) => {
+    return {...state, ...{ inflight: false, error: action.error }};
```
chuckwondo: nit: no need for using spread operator on those trailing elements; just put them directly into the new object

app/src/js/reducers/granules.js
```diff
@@ -65,141 +66,157 @@ export const initialState = {
...
+
+  [OPTIONS_COLLECTIONNAME]: (state, action) => {
+    const { data } = action;
+    // Map the list response to an object with key-value pairs like:
+    // displayValue: optionElementValue
+    const options = data.results.reduce((obj, d) => {
+      const { name, version } = d;
+      obj[`${name} ${version}`] = `${name}___${version}`;
+      return obj;
+    }, {});
```
chuckwondo: nit: replace `${name}___${version}` with `getCollectionId(d)` (from `app/src/js/utils/format.js`)

or, if you're feeling a bit more ambitious, change the highlighted block to this (although this is arguably no more readable, so feel free to ignore my digression):
```
const options = action.data.results.reduce(
  (obj, { name, version }) =>
    Object.assign(obj, {
      [`${name} ${version}`]: getCollectionId({ name, version })
    }),
  {}
);
```

app/src/js/reducers/mmtLinks.js
```diff
@@ -5,17 +5,14 @@ import { getCollectionId } from '../utils/format';
...
-      const collectionId = getCollectionId(data);
-      set(state, [collectionId], data.url);
-      break;
-  }
-  return state;
-}
+export default createReducer(initialState, {
+  [ADD_MMTLINK]: (state, action) => {
+    const { data } = action;
+    const collectionId = getCollectionId(data);
```
chuckwondo: Aha! See `getCollectionId` is being used here (as I suggested elsewhere). :-)

app/src/js/reducers/schema.js
```diff
@@ -3,18 +3,17 @@
...
-  switch (type) {
-    case SCHEMA:
-      return Object.assign({}, state, { [key(config.url)]: action.data });
-    default:
-      return state;
+export default createReducer(initialState, {
+
+  [SCHEMA]: (state, action) => {
+    const { config } = action;
+    return {...state, ...{ [key(config.url)]: action.data }};
```
chuckwondo: nit: no need for second spread; just do this:
```
return { ...state, [key(config.url)]: action.data };
```

cypress/integration/providers_spec.js
chuckwondo: Does this cause the user any problem?

test/reducers/stats.js
```diff
@@ -0,0 +1,319 @@
...
+
+  const expected = {
+    ...testState,
+    ...{
+      stats: {
+        data: { ...data, ...{ queriedAt: new Date(testStart) } },
+        inflight: false,
+        error: null
+      }
+    }
```
chuckwondo: nit: no need for extra nesting and spread operator

test/reducers/stats.js
```diff
@@ -0,0 +1,319 @@
...
+});
+
+test('stats_inflight', (t) => {
+  const testState = cloneDeep(initialState);
+
+  const action = { type: STATS_INFLIGHT };
+
+  const expected = {
+    ...testState,
+    ...{ stats: { ...testState.stats, ...{ inflight: true } } }
```
chuckwondo: nit: no need for extra nesting and spread operator (2 places on this line)

test/reducers/stats.js
```diff
@@ -0,0 +1,319 @@
...
+
+  const expected = {
+    ...initialState,
+    ...{
+      stats: {
+        ...initialState.stats,
+        ...{ inflight: false },
+        ...{ error: 'an error' }
+      }
+    }
```
chuckwondo: nit: no need for the extra nestings and spreadings

cypress/integration/providers_spec.js
dopeters: Couldn't resist taking a peek at this.

Looks like the issue is that in `componentDidUpdate` of `EditRecord`, it is grabbing `state` from `prevProps` instead of `this.props`. Now that we are rendering far fewer times, the prevProps correctly does not have a success status. Pulling the state out of this.props seems to fix it.

cypress/integration/providers_spec.js
chuckwondo: Not sure what that's referring to, but please create a new issue, if it makes sense to do so.

cypress/integration/providers_spec.js
chuckwondo: Sweet. Nice find. @flamingbear do you care if @dopeters pushes this tweak?

flamingbear: Sorry I get my emails slowly.  I chatted and @dopeters is pushing to a branch for me.

flamingbear: probably overkill

dopeters: Added fix here https://github.com/nasa/cumulus-dashboard/pull/704

app/src/js/reducers/collections.js
flamingbear: So @cumulus/api is only a dev dependency, not a runtime dependency.   But I will use the add that function to our format library.

flamingbear:
> My only request is that you please create a new issue (as discussed in our Architecture meeting) for tidying up all of the `createReducer` occurrences with a consistent form of "state mutation" to avoid potential confusion in the future (about why there are inconsistencies), along with perhaps showing examples of different options

https://bugs.earthdata.nasa.gov/browse/CUMULUS-1810

PR #707, opened 2020-03-27
[CUMULUS-1509] Update granule page styles

dopeters:
Updates the granules pages

* Added breadcrumbs
* Moved "Run bulk granules" into bulk actions and reworked that functionality
* Remved `Granules/index.js` re-dispatching listGranules at its own interval
* Updated labels/placeholders for list filters


dopeters:
> Looks great! Love that we have implemented labels for the inputs. The refining of grouping all the CTAs in one place was needed. Thank you.
> 
> I just have one thing that isn't a big deal but it stands out as far as spacing between the heading and the search & filters area. I think we could change the spacing for `.heading__wrapper--border` from `margin-bottom: 2em `to `margin-bottom: 0em`. It might be something we will need to come back to once we remove the 'next updates' area from all pages since the border would be an issue for 'Workflows Overview' page. Maybe can try to get it as close to 0 as possible to make the section look cohesive but if it is a problem for other pages then we can wait.

Ending up reducing `margin-bottom` to 1em, to avoid completely removing the spacing from some of the pages without tables. Also updated the last of the tables that were not using the <ListFilter> component due to some funkiness with the spacing

PR #708, opened 2020-03-30
Cumulus 1527

jjmccoy:
Style how individual rule page will look and feel. Add the state "enabled" and "disabled" badge indicators and restyle the details table. The sidebar does not contain "enabled" and "disabled" because we do not have those subpages built out yet. Currently, working on the logic for showing "enabled" and "disabled" badges.

### DIFF
--- a/app/src/js/components/Rules/rule.js
+++ b/app/src/js/components/Rules/rule.js
@@ -27,6 +27,22 @@ import Loading from '../LoadingIndicator/loading-indicator';
 import Metadata from '../Table/Metadata';
 import AsyncCommands from '../DropDown/dropdown-async-command';
 import ErrorReport from '../Errors/report';
+import Breadcrumbs from '../Breadcrumbs/Breadcrumbs';
+
+const breadcrumbConfig = [
+  {
+    label: 'Dashboard Home',
+    href: '/'
+  },
+  {
+    label: 'Rules',
+    href: '/Rules'
+  },
+  {
+    label: 'Rule Overview',
+    active: true
+  }
+];
 
 const metaAccessors = [
   ['RuleName', 'name'],
@@ -157,6 +173,11 @@ class Rule extends React.Component {
     const errors = this.errors();
     return (
       <div className='page__component'>
+        <section className='page__section page__section__controls'>
+          <div className="options--top">
+            <Breadcrumbs config={breadcrumbConfig} />
+          </div>
+        </section>
         <section className='page__section page__section__header-wrapper'>
           <div className='page__section__header'>
             <h1 className='heading--large heading--shared-content with-description'>{ruleName}</h1>
@@ -174,13 +195,6 @@ class Rule extends React.Component {
               className='button button--edit button--small button--green form-group__element--right'
               to={`/rules/edit/${ruleName}`}>Edit</Link>
             {lastUpdated(data.timestamp)}
-            {data.state ? (
-              <dl className='status--process'>
-                <dt>State:</dt>
-                <dd className={data.state.toLowerCase()}>{displayCase(data.state)}</dd>
-              </dl>
-            ) : null}
-
           </div>
         </section>
         <section className='page__section'>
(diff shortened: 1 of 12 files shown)


app/src/js/components/Table/Metadata.js
```diff
@@ -15,8 +15,10 @@ class Metadata extends React.Component {
             value = meta[2](value);
           }
           return acc.concat([
-            <dt key={`meta-${meta[1]}--dt`}>{meta[0]}</dt>,
-            <dd key={`meta-${meta[1]}--dd`}>{value}</dd>
+            <div className="meta__row">
+              <dt key={`meta-${meta[1]}--dt`}>{meta[0]}</dt>
+              <dd key={`meta-${meta[1]}--dd`}>{value}</dd>
+            </div>
```
dopeters: Looks like this change is causing a test failure in `test/components/reconciliation-reports/reconciliation-report.js`. In the test "show individual report", we have

```
  // ReconciliationReport is configured to use 6 metaAccessors,
  // so there will be 6 groups of dt, dd elements for a total of 12
  t.is(MetadataWrapperChildren.length, 12);
```

Since the dt and dd elements are now grouped, we should be able to do 
`  t.is(MetadataWrapperChildren.length, 6);` instead.

app/src/js/components/Rules/rule.js
```diff
@@ -169,26 +190,29 @@ class Rule extends React.Component {
...
           {errors.length ? <ErrorReport report={errors} /> : null}
           <div className='heading__wrapper--border'>
             <h2 className='heading--medium with-description'>Rule Overview</h2>
           </div>
-          <Metadata data={data} accessors={metaAccessors} />
+          <div className="rule__state">
+            {data.state ? (
+              <dl className='status--process'>
+                <dt>State:</dt>
+                <dd className={`status--badge status--badge__${data.state.toUpperCase()}`}>{displayCase(data.state)}</dd>
```
dopeters: Curious why this is `toUpperCase` instead of `toLowerCase`. Currently when the rule is disabled, the badge is still green.

app/src/js/components/Rules/rule.js
```diff
@@ -27,15 +27,31 @@ import Loading from '../LoadingIndicator/loading-indicator';
...
+  {
+    label: 'Rule Overview',
+    active: true
+  }
+];
 
 const metaAccessors = [
-  ['RuleName', 'name'],
+  ['Rule Name', 'name'],
+  ['Timestamp', 'timestamp', fullDate],
```
dopeters: Wondering if instead of timestamp we want to be looking at `createdAt` and/or `updatedAt` since it seems like timestamp is empty.

app/src/js/components/Rules/rule.js
```diff
@@ -27,15 +27,31 @@ import Loading from '../LoadingIndicator/loading-indicator';
...
   ['Workflow', 'workflow'],
```
dopeters: Also just wondering if we should try to link to the workflow like we link to the provider

app/src/js/components/Table/Metadata.js
jjmccoy: Thanks for helping with the test failure. I notices it's failing in reports since I added a div group but was trying to figure out how to fix it. Much appreciated.

app/src/js/components/Rules/rule.js
jjmccoy: The `metaAccessors` code was originally there. I was just making sure things were spelled right or the content made sense. I think `updatedAt` would be good since it is existing to add to the implementation ticket.

app/src/js/components/Rules/rule.js
jjmccoy: Sorry again. This is a style ticket. There might be an implementation ticket which I would suggest to add this.

app/src/js/components/Rules/rule.js
dopeters: Fair enough, makes sense!

app/src/js/components/Rules/rule.js
jjmccoy: - Sorry for any confusion. I have designed all badges to be upper case based off of our EED EUI components (https://cdn.earthdata.nasa.gov/eui/latest/docs/eui/index.html#elements). 
- Oops. I left the test in there. Let me get that. Thanks for catching that.

dopeters: Does this apply to the class name itself though? Now I'm not seeing a background color for either state due to case sensitivity of the class selector.

jjmccoy: Oh I see what you are saying. Nevermind. Silly me. I was thinking about the badge itself and not className itself.

dopeters: Cool no worries, looks great! Thanks for the changes

PR #712, opened 2020-04-02
CUMULUS-1528

chuckwondo:
Changed "Add Rule" form from raw JSON input to form with individual fields.

### DIFF
--- a/app/src/js/components/Form/Form.js
+++ b/app/src/js/components/Form/Form.js
@@ -267,7 +267,7 @@ export class Form extends React.Component {
 
         {this.props.submit && (
           <button
-            className={'button button__animation--md button__arrow button__arrow--md button__animation button__arrow--white button__submit' + (this.isInflight() ? ' button--disabled' : '')}
+            className={'button button__animation--md button__arrow button__arrow--md button__animation button__arrow--white button--submit' + (this.isInflight() ? ' button--disabled' : '')}
             onClick={this.onSubmit}
           >
             {this.isInflight() ? 'Loading...' : status === 'success' ? 'Success!' : 'Submit'}
@@ -276,7 +276,7 @@ export class Form extends React.Component {
 
         {this.props.cancel && (
           <button
-            className={'button button__animation--md button__arrow button__arrow--md button__animation button--secondary form-group__element--left button__cancel' + (this.isInflight() ? ' button--disabled' : '')}
+            className={'button button__animation--md button__arrow button__arrow--md button__animation button--secondary form-group__element--left button--cancel' + (this.isInflight() ? ' button--disabled' : '')}
             onClick={this.onCancel}
           >
             Cancel
--- a/app/src/js/components/Form/Form.js
+++ b/app/src/js/components/Form/Form.js
@@ -1,5 +1,7 @@
 'use strict';
+
 import React from 'react';
+import { createNextState } from '@reduxjs/toolkit';
 import { generate } from 'shortid';
 import { set } from 'object-path';
 import slugify from 'slugify';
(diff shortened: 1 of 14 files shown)


app/src/js/components/FormSchema/schema.js
```diff
@@ -36,65 +48,91 @@ export const traverseSchema = function (schema, fn, path) {
...
+      schemaProperty.required.includes(property);
 
+    const labelText = meta.title || property;
     const label = (
       <span>
-        { path ? <span className='label__path'>{path + ' - '}</span> : null }
-        <span className='label__name'>{meta.title || property}</span>
-        { required ? <span className='label__required'> *</span> : null }
-        { meta.description ? <span className='label__description'> ({meta.description})</span> : null }
+        <span className="label__name">{labelText}</span>
```
dopeters, line 111: While we're in here, any chance we can call `displayCase` on the `labelText` to get consistent capitalization?

app/src/js/components/Form/Form.js
```diff
@@ -42,147 +76,139 @@ const errorMessage = (errors) => `Please review ${errors.join(', ')} and submit
...
+        // Ignore empty fields that aren't required
+        // These may have input elements in the form, but
+        // the API will fail if it's sent empty strings
+        if (value !== '' || field.required) {
+          // Use 'set' function since the value of 'input.schemaProperty' may
+          // be a "path", such as 'rule.type', so we would want to set a nested
+          // property (i.e., the 'type' property within a 'rule' property).
+          set(payload, field.schemaProperty, value);
+        }
+      });
```
dopeters, line 168: General form comment so not necessarily something we should worry about here, but something's up with required fields. I would expect the first check the form does is that all required fields are filled. If I submit a completely empty form, there's no user feedback. Then, if I fill out the form, only some of the required fields will show the required error vs while others will have an error from the API. We should probably add a ticket to deal with the form logic/error handling/validation better

app/src/js/components/Rules/add.js
```diff
@@ -1,99 +1,158 @@
...
-      pk={'new-rule'}
-      title={title}
+    <AddRecord
+      data={rule}
+      schemaKey={'rule'}
       primaryProperty={'name'}
+      title={title}
       state={rules}
-      defaultValue={defaultValue}
+      baseRoute={'/rules/rule'}
```
dopeters, line 133: Somethings up here, because clicking Cancel brings you back to `/rules/rules`, which is a blank page

app/src/js/components/Form/Form.js
```diff
@@ -202,22 +228,23 @@ export class Form extends React.Component {
...
 
-            // retrieve value and errors stored in state
-            let inputId = this.generateComponentId(form.schemaProperty);
+            let inputId = generateComponentId(input.schemaProperty, this.id);
             let { value, error } = inputState[inputId];
+
             // coerce non-null values to string to simplify proptype warnings on numbers
             if (type !== formTypes.list && type !== formTypes.subform && !value && value !== 0) {
               value = String(value);
             }
```
dopeters, line 237: Adding this comment here because there's not really a better place for it

Not related to this PR, but the optional tag field makes it really noticeable. Our `button--disabled` class isn't doing what it should. I traced it back to line 25 in `_buttons.scss`

```
  &:disabled {
    opacity: 0.6;
    pointer-events: none;
  }
```

We're using the pseudo selector and not the defined class. 

We could fix this by adding the class as a selector like this:

```
  &:disabled,
  &--disabled {
    opacity: 0.6;
    pointer-events: none;
  }
```

We can also just add this change to a future design ticket, but I think it would make that tag field a little more clear.

app/src/js/components/Form/Form.js
chuckwondo: Agreed. There are actually a number of situations where I've seen errors occur silently without indication to the user that something went wrong. Matt and I also discussed this the other day, so I'll create a ticket for it.

app/src/js/components/Rules/add.js
```diff
@@ -1,99 +1,158 @@
...
       createRecord={createRule}
-      getBaseRoute={getBaseRoute}
-      getPk={getRuleName}
-      requireConfirmation={true}
-      type={'rule'}
-      ModalBody={ModalBody}
```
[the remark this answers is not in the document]
chuckwondo: @jjmccoy Can you weigh in here?

PR #713, opened 2020-04-02
Cumulus 1515 workflow search filter

flamingbear:
CUMULUS-1515

Implements workflow filter on overview page.
Refactors the workflow shape to be more intutive.
Updates the WORKFLOW reducer to filter based on the searchString.
Adds unit tests for reducer.
Adds cypress tests to display behavior.

Bonus: Creates a link to the stepfunction. (needs to have AWS_REGION included somehow)
Bonus: Adds Steps on the workflow page. (maybe this is trash?)


app/src/js/config/config.js
```diff
@@ -9,6 +9,7 @@ const config = {
     }
   },
   apiRoot: process.env.APIROOT || 'https://example.com',
+  awsRegion: process.env.AWS_REGION || 'us-west-2',
```
kkelly51, line 12: ummm... I don't think we can assume this is the default to use. Does it matter what is here?

app/src/js/reducers/workflows.js
```diff
@@ -3,19 +3,18 @@ import { set } from 'object-path';
...
 export const initialState = {
-  list: {
-    data: [],
-    meta: {},
-    params: {},
-    inflight: false,
-    error: false
-  },
-  map: {}
+  data: [],
```
kkelly51, line 13: Why did you want to get rid of list?

app/src/js/config/config.js
flamingbear: Why can't we assume that as a default?  A user would just override with whatever they have set.  And I'm pretty sure most daacs are being launched into us-west-2

app/src/js/reducers/workflows.js
flamingbear: Why did we want to keep it?
It made for a deep structure and was somewhat confusing.  I thought flattening was just simplification.

app/src/js/reducers/workflows.js
```diff
@@ -3,19 +3,18 @@ import { set } from 'object-path';
...
 
 export const initialState = {
-  list: {
-    data: [],
-    meta: {},
-    params: {},
-    inflight: false,
-    error: false
-  },
-  map: {}
```
kkelly51, line 18: What was this map here for? Nothing? Should we add it back?

flamingbear: Good catch, it should be there.  It gets filled on the first call to the WORKFLOWS reducer

cypress/integration/workflows_spec.js
```diff
@@ -52,5 +55,15 @@ describe('Dashboard Workflows Page', () => {
...
     });
+
+    it('filters workflows when a user types in the search box', () => {
+      cy.server();
+      cy.route('GET', '/workflows*').as('get-workflows');
+      cy.visit('/workflows');
+      cy.get('.table .tbody .tr').its('length').should('be.eq', 2);
+      cy.get('.table .tbody .tr').first().contains('HelloWorldWorkflow');
+      cy.get('.search').click().type('cond');
+      cy.get('.table .tbody .tr').first().contains('SecondTestWorkflow');
```
kkelly51, line 66: Can we add one more line to this test checking that that length has gone down to 1?

cypress/integration/workflows_spec.js
```diff
@@ -52,5 +55,15 @@ describe('Dashboard Workflows Page', () => {
...
       });
     });
+
+    it('filters workflows when a user types in the search box', () => {
+      cy.server();
+      cy.route('GET', '/workflows*').as('get-workflows');
+      cy.visit('/workflows');
+      cy.get('.table .tbody .tr').its('length').should('be.eq', 2);
+      cy.get('.table .tbody .tr').first().contains('HelloWorldWorkflow');
+      cy.get('.search').click().type('cond');
```
kkelly51, line 65: love that this can be in the middle of the name instead of starting at the beginning

app/src/js/reducers/workflows.js
kkelly51: I guess if there's no reason to keep it. I just thought there was a reason all the reducers have the list structure. Did it have to do with going into Tables? or Lists?

flamingbear: I did not need it.  And it makes the reducers slightly more complicated, so I got rid of it.  Easy enough to add back if we find it useful. But I'd prefer to wait unless we know a reason for the structure.

app/src/js/reducers/workflows.js
```diff
@@ -26,20 +25,41 @@ function createMap (data) {
...
+/**
+ * Filters intput data by filterstring on the data'a object's name.
+*
+ * @param {Array} rawData - An array of objects with a name parameter.
+ * @param {string} filterString - a string to check if name includes.
+ * @returns {Array} Filtered Array of rawData's objects who's names include filterString.
+ */
+export const filterData = (rawData, filterString) => {
+  if (filterString !== null) {
+    return rawData.filter(d => d.name.includes(filterString));
```
kkelly51, line 37: could we make this case insensitive?

cypress/integration/workflows_spec.js
flamingbear: Sure.

cypress/integration/workflows_spec.js
flamingbear: It is the middle.  the string is "SecondTestWorkflow", so 'cond', but I will change it to 
'conndtes' to make sure it's case insensitve.

app/src/js/reducers/workflows.js
flamingbear: yes.

app/src/js/reducers/workflows.js
flamingbear: 7ef28d16

cypress/integration/workflows_spec.js
flamingbear: 7ef28d16

app/src/js/reducers/workflows.js
flamingbear: 7ef28d16

cypress/integration/workflows_spec.js
flamingbear: 7ef28d16

app/src/js/reducers/workflows.js
flamingbear: I'm looking at this. And there's reasons, but they're not documented, and I don't fully understand them, and they don't work for workflow. I will ping you separately.

flamingbear: I will add it back in this case.  There's a bunch of hidden assumptions for sure.

flamingbear: d65552cb

PR #714, opened 2020-04-04
Fix selector in spec to remove wait().

flamingbear:
I'm pretty certain, this link below is describing a bunch of problems we're
having.  I have been unable to recreate the failed test since switching the get
selector.

Everyone should read this whole thing, but the TL;DR:

"For a variety of implementation reasons, Cypress commands only retry the last
command before the assertion."

https://docs.cypress.io/guides/core-concepts/retry-ability.html#Only-the-last-command-is-retried


flamingbear:
this is included in my other pr

flamingbear:
@bt6453 This is OBE now, because I merged in a fix for everywhere we were doing that in #715 
so you might just check out that PR for background knowledge

PR #715, opened 2020-04-04
Fixes "obvious" selector problems.

flamingbear:
I'm pretty certain, this link below is describing a bunch of problems we're
having.

You should read this whole thing, but the TL;DR:

"For a variety of implementation reasons, Cypress commands only retry the last
command before the assertion."

https://docs.cypress.io/guides/core-concepts/retry-ability.html#Only-the-last-command-is-retried

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -3,5 +3,7 @@
   "env": {
     "APIROOT": "http://localhost:5001"
   },
-  "nodeVersion": "system"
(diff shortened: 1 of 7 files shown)


flamingbear:
@dopeters Thanks for review.  I'm going to drop that link in chat for everyone.  And yes we need to be careful about our tests.  I think we're getting better.

PR #716, opened 2020-04-05
Quick work around for rules input bug.

flamingbear:
So I'm pretty convinced we are typing in the input field before it's ready to type.  

I found this looking for the bug and 
https://www.cypress.io/blog/2018/02/05/when-can-the-test-start/

I'd like to eventually try to implement spying on the dom as suggested in the article, but this seems to stop the bleeding.  I left comments in the spec pointing to the bug and the above article so we can come back to this sometime.

You can see in the image on the failed test. the rule name is `ewRule` 
Screen Shot 2020-04-05 at 11 35 37 AM

### DIFF
--- a/cypress/integration/rules_spec.js
+++ b/cypress/integration/rules_spec.js
@@ -106,44 +106,47 @@ describe('Rules page', () => {
       const collection = getCollectionId({ name: 'MOD09GQ', version: '006' });
 
       // Fill the form and submit
+      // mhs: I think we're seeing https://www.cypress.io/blog/2018/02/05/when-can-the-test-start/
+      // Typing into a form that's not ready. (https://github.com/cypress-io/cypress/issues/3817)
       cy.get('form div ul').as('ruleInput');
       cy.get('@ruleInput')
         .contains('name', { matchCase: false })
         .siblings('input')
+        .wait(500)
         .type(ruleName);
       cy.get('@ruleInput')
         .contains('.dropdown__label', 'workflow', { matchCase: false })
         .siblings()
         .find('select')
-        .select(workflow, { force: true })
+        .select(workflow)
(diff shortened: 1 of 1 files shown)


cypress/integration/rules_spec.js
```diff
@@ -106,44 +106,47 @@ describe('Rules page', () => {
       const collection = getCollectionId({ name: 'MOD09GQ', version: '006' });
 
       // Fill the form and submit
+      // mhs: I think we're seeing https://www.cypress.io/blog/2018/02/05/when-can-the-test-start/
+      // Typing into a form that's not ready. (https://github.com/cypress-io/cypress/issues/3817)
       cy.get('form div ul').as('ruleInput');
       cy.get('@ruleInput')
         .contains('name', { matchCase: false })
         .siblings('input')
+        .wait(500)
```
flamingbear, line 115: this is the workaround, that we don't want to rely on...

cypress/integration/rules_spec.js
```diff
@@ -106,44 +106,47 @@ describe('Rules page', () => {
...
         .contains('.dropdown__label', 'state', { matchCase: false })
         .siblings()
         .find('select')
-        .select('ENABLED', { force: true })
+        .select('ENABLED')
         .should('have.value', 'ENABLED');
 
       cy.contains('form button', 'Submit').click();
-
+      cy.url().should('include', 'rule/newRule');
```
flamingbear, line 149: We probably don't need this, but it will retry until we reach that page.  I expect it's probably the same-ish test as L 151.

flamingbear:
@chuckwondo this is OBE by my pulling in the eslint update, but you should still take a look if you get a chance.

flamingbear:
Yup, Thanks for checking it out.

PR #719, opened 2020-04-08
Update api version to v1.21.0

flamingbear:
This combination of package-lock.json appears to work fine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1169,22 +1169,23 @@
       "dev": true
     },
     "@cumulus/api": {
-      "version": "1.20.0",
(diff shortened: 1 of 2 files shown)


flamingbear:
This PR closes CUMULUS-1824, but does not unblock @dopeters 

I've tried to update the @cumulus/api to 1.21.1-alpha.0, but that didn't have the correct dependency on @cumulus/common, so I published a common alpha and am coming back to see if I can't get the api to use it correctly.

flamingbear:
I'm going to close this because I think I can update to 1.12.1-alpha.1 now.

PR #722, opened 2020-04-09
[CUMULUS-1813] Fix execution graph on Execution page

bt6453:
So Far:

* Remove datepicker on execution overview and logs page

### DIFF
--- a/app/src/js/components/Executions/execution-status-graph.js
+++ b/app/src/js/components/Executions/execution-status-graph.js
@@ -35,7 +35,9 @@ class ExecutionStatusGraph extends React.Component {
     var svg = d3.select('svg');
     render(svg, this.g);
     var height = d3.select('svg g').node().getBBox().height;
+    var width = d3.select('svg g').node().getBBox().width;
     svg.style('height', height + 10);
+    svg.style('width', width);
     svg.style('padding-right', 150);
   }
 
--- a/app/src/js/components/Executions/execution-status-graph.js
+++ b/app/src/js/components/Executions/execution-status-graph.js
@@ -36,9 +36,11 @@ class ExecutionStatusGraph extends React.Component {
     render(svg, this.g);
     var height = d3.select('svg g').node().getBBox().height;
     var width = d3.select('svg g').node().getBBox().width;
-    svg.style('height', height + 10);
-    svg.style('width', width);
-    svg.style('padding-right', 150);
+    console.log('HEIGHT: ' + height);
+    console.log('WIDTH: ' + width);
+    svg.attr('height', height + 10);
+    svg.attr('width', width);
+    svg.attr('padding-right', 150);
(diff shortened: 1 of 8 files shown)


app/src/js/components/Executions/execution-status-graph.js
```diff
@@ -27,16 +27,21 @@ class ExecutionStatusGraph extends React.Component {
 
     const workflow = JSON.parse(stateMachine.definition);
 
+    console.log('WORKFLOW: ' + JSON.stringify(workflow));
+
     var events = getExecutionEvents(executionHistory);
+
     var graph = workflowToGraph(workflow);
+    console.log('GRAPH: ' + JSON.stringify(graph));
```
dopeters, line 35: Do we still need these console logs?

app/src/js/components/Executions/index.js
```diff
@@ -21,10 +21,27 @@ class Executions extends React.Component {
...
+        </div>
+      );
+    }
+  }
+
   render () {
     return (
       <div className='page__workflows'>
-        <DatePickerHeader onChange={this.query} heading={strings.executions}/>
+        {this.renderHeader()}
```
dopeters, line 44: Just curious why we're conditionally displaying `DatePickerHeader` here? We don't do this for an individual collection page, so not sure why we're doing it here.

app/src/js/components/Executions/execution-status-graph.js
```diff
@@ -27,16 +27,21 @@ class ExecutionStatusGraph extends React.Component {
...
     addEventsToGraph(events, graph);
     this.g = draw(graph);
     var render = new dagre.render();
     var svg = d3.select('svg');
```
dopeters, line 39: Fwiw, when I was working on my ticket, I noticed that since this is selecting 'svg', it grabs the first one it finds. When the datepicker is present, the svg within the start date is what is selected and causes the weird styling. If we add a class to the svg we're rendering here and then select for that, it should solve the issue. If that's the reason we're hiding the datepicker header, than we can avoid that by doing this 

```
var svg = d3.select('.visual');
.
.
.
.
render () {
  return (
    <svg className='visual'></svg>
  );
}
```

app/src/js/components/Executions/execution-status-graph.js
bt6453: Good catch. I will remove these.

app/src/js/components/Executions/index.js
bt6453: So in discussions about this when I started the ticket, the DatePicker doesn't actually affect anything on the individual execution status page, and it was suggested that it didn't need to be on the page. So I removed it.

dopeters: Gotcha, that's fair enough

PR #726, opened 2020-04-14
Fix listCollections dispatch on Add Rules page

dopeters:
1. On add rules page, we are not checking for errors when determining if an action has been dispatches which can lead to endless rerender. Added that logic and make object destructing more clear.

2. We need a way to retrieve all collections on the add rules page for populating the dropdown. (if a time filter is applied, only collections with active granules will populate that dropdown). Added a `listAll` property to the options passed to `listCollections`


app/src/js/actions/index.js
```diff
@@ -128,16 +128,17 @@ export const checkApiVersion = () => {
 };
 
 export const listCollections = (options) => {
+  const { listAll, ...queryOptions } = options;
   return (dispatch, getState) => {
     const timeFilters = fetchCurrentTimeFilters(getState().datepicker);
-    const urlPath = `collections${isEmpty(timeFilters) ? '' : '/active'}`;
+    const urlPath = `collections${isEmpty(timeFilters) || listAll ? '' : '/active'}`;
```
dopeters, line 134: Can test that the endless rerender is no longer happening by changing `/active` here to something that doesn't exist

flamingbear:
I know it must have been a different PR, but will you pin the @cumulus/api to exactly 1.21.1-alpha.2? instead of ^1.21.1-alpha.2.  We will thank ourselves later I'm sure.

app/src/js/actions/index.js
flamingbear: I was able to install 1.21.1-alpha.1 and verify this no longer occurs as well.

app/src/js/components/Rules/add.js
```diff
@@ -108,11 +108,14 @@ const AddRule = ({
     }
   }, [name, rulesMap, isCopy]);
 
-  const dispatched = ({ list: { inflight, meta: { queriedAt } } }) =>
-    inflight || queriedAt !== undefined;
+  const dispatched = ({ list }) => {
+    const { inflight, meta, error } = list || {};
+    const { queriedAt } = meta || {};
+    return error || inflight || queriedAt !== undefined;
+  };
```
flamingbear, line 115: 💯

PR #729, opened 2020-04-15
CUMULUS-1787 collections header

dopeters:
1. When time filter is applied, header on collections page says "Active Collections". Otherwise, it says "All Collections". Also cleaned up some extraneous code on that page.

2. Removes time filter from the `listCollections` and `listProviders` actions on the Add Rule page by passing in `listAll` option

Can test by setting date to a time after data was seeded/ingested. When adding a rule, all collections and providers should still be populated. Did not change the workflows action since it does not take in a timefilter.

### DIFF
--- a/app/src/js/actions/index.js
+++ b/app/src/js/actions/index.js
@@ -128,9 +128,9 @@ export const checkApiVersion = () => {
 };
 
 export const listCollections = (options = {}) => {
-  const { listAll, ...queryOptions } = options;
+  const { listAll = false, getMMT = true, ...queryOptions } = options;
   return (dispatch, getState) => {
-    const timeFilters = fetchCurrentTimeFilters(getState().datepicker);
+    const timeFilters = !listAll ? fetchCurrentTimeFilters(getState().datepicker) : {};
     const urlPath = `collections${isEmpty(timeFilters) || listAll ? '' : '/active'}`;
     return dispatch({
       [CALL_API]: {
@@ -140,7 +140,11 @@ export const listCollections = (options = {}) => {
         url: new URL(urlPath, root).href,
         qs: Object.assign({ limit: defaultPageLimit }, queryOptions, timeFilters)
       }
-    }).then(() => dispatch(getMMTLinks()));
+    }).then(() => {
+      if (getMMT) {
+        dispatch(getMMTLinks());
+      }
+    });
   };
 };
 
@@ -582,15 +586,16 @@ export const clearPdrsSearch = () => ({ type: types.CLEAR_PDRS_SEARCH });
 export const filterPdrs = (param) => ({ type: types.FILTER_PDRS, param: param });
 export const clearPdrsFilter = (paramKey) => ({ type: types.CLEAR_PDRS_FILTER, paramKey: paramKey });
 
-export const listProviders = (options) => {
+export const listProviders = (options = {}) => {
+  const { listAll = false, queryOptions } = options;
   return (dispatch, getState) => {
-    const timeFilters = fetchCurrentTimeFilters(getState().datepicker);
+    const timeFilters = !listAll ? fetchCurrentTimeFilters(getState().datepicker) : {};
     return dispatch({
       [CALL_API]: {
         type: types.PROVIDERS,
         method: 'GET',
         url: new URL('providers', root).href,
-        qs: Object.assign({ limit: defaultPageLimit }, options, timeFilters)
+        qs: Object.assign({ limit: defaultPageLimit }, queryOptions, timeFilters)
       }
     });
   };
--- a/app/src/js/actions/index.js
+++ b/app/src/js/actions/index.js
@@ -130,7 +130,7 @@ export const checkApiVersion = () => {
 export const listCollections = (options = {}) => {
   const { listAll = false, getMMT = true, ...queryOptions } = options;
   return (dispatch, getState) => {
-    const timeFilters = !listAll ? fetchCurrentTimeFilters(getState().datepicker) : {};
+    const timeFilters = listAll ? {} : fetchCurrentTimeFilters(getState().datepicker);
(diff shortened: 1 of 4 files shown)


app/src/js/actions/index.js
```diff
@@ -128,9 +128,9 @@ export const checkApiVersion = () => {
 };
 
 export const listCollections = (options = {}) => {
-  const { listAll, ...queryOptions } = options;
+  const { listAll = false, getMMT = true, ...queryOptions } = options;
   return (dispatch, getState) => {
-    const timeFilters = fetchCurrentTimeFilters(getState().datepicker);
+    const timeFilters = !listAll ? fetchCurrentTimeFilters(getState().datepicker) : {};
```
chuckwondo, line 133: nit: I would flip this to avoid the negation:
```javascript
const timeFilters = listAll ? {} : fetchCurrentTimeFilters(getState().datepicker);
```

app/src/js/actions/index.js
```diff
@@ -582,15 +586,16 @@ export const clearPdrsSearch = () => ({ type: types.CLEAR_PDRS_SEARCH });
 export const filterPdrs = (param) => ({ type: types.FILTER_PDRS, param: param });
 export const clearPdrsFilter = (paramKey) => ({ type: types.CLEAR_PDRS_FILTER, paramKey: paramKey });
 
-export const listProviders = (options) => {
+export const listProviders = (options = {}) => {
+  const { listAll = false, queryOptions } = options;
   return (dispatch, getState) => {
-    const timeFilters = fetchCurrentTimeFilters(getState().datepicker);
+    const timeFilters = !listAll ? fetchCurrentTimeFilters(getState().datepicker) : {};
```
chuckwondo, line 592: nit: same here for removing the negation

app/src/js/actions/index.js
```diff
@@ -582,15 +586,16 @@ export const clearPdrsSearch = () => ({ type: types.CLEAR_PDRS_SEARCH });
 export const filterPdrs = (param) => ({ type: types.FILTER_PDRS, param: param });
 export const clearPdrsFilter = (paramKey) => ({ type: types.CLEAR_PDRS_FILTER, paramKey: paramKey });
 
-export const listProviders = (options) => {
+export const listProviders = (options = {}) => {
+  const { listAll = false, queryOptions } = options;
```
chuckwondo, line 590: I noticed that this is _not_ using the rest params notation, but `app/src/js/actions/index.js` _is_ using rest params notation. In other words, compare this line from the other file to the line above:
```javascript
const { listAll = false, getMMT = true, ...queryOptions } = options;
```
However, both _use_ the `queryOptions` parameter afterwards in the same way. I suspect that one of these is erroneous.

app/src/js/components/Collections/list.js
```diff
@@ -103,7 +98,10 @@ class CollectionList extends React.Component {
         </section>
         <section className='page__section'>
           <div className='heading__wrapper--border'>
-            <h2 className='heading--medium heading--shared-content with-description'>{strings.all_collections} <span className='num--title'>{count ? ` ${tally(count)}` : 0}</span></h2>
+            <h2 className='heading--medium heading--shared-content with-description'>
+              {hasTimeFilter ? 'Active Collections' : strings.all_collections}
```
chuckwondo, line 102: nit: consider adding `'Active Collections'` to the `strings` object for consistency.

app/src/js/components/Collections/list.js
```diff
@@ -103,7 +98,10 @@ class CollectionList extends React.Component {
...
+              <span className='num--title'>{count ? ` ${tally(count)}` : 0}</span>
```
chuckwondo, line 103: Is there a reason that we need a space preceding the tally when the count is non-zero, but not when it is zero?

app/src/js/components/Rules/add.js
```diff
@@ -115,8 +115,8 @@ const AddRule = ({
   };
 
   useEffect(() => {
-    if (!dispatched(collections)) dispatch(listCollections({ listAll: true }));
-    if (!dispatched(providers)) dispatch(listProviders());
+    if (!dispatched(collections)) dispatch(listCollections({ listAll: true, getMMT: false }));
+    if (!dispatched(providers)) dispatch(listProviders({ listAll: true }));
```
chuckwondo, line 119: Excellent!

app/src/js/actions/index.js
dopeters: Yep, I was moving some stuff around and this got messed up in the shuffle. Nice catch!

app/src/js/components/Collections/list.js
dopeters: Doesn't seem to be a good reason. Will remove

app/src/js/components/Collections/list.js
dopeters: Sure thing

app/src/js/actions/index.js
dopeters: Yeah I like this better. Done.

dopeters:
Thanks for the thorough review @chuckwondo! Lots of good suggestions, all of which I incorporated.

app/src/js/components/Collections/list.js
```diff
@@ -103,7 +98,10 @@ class CollectionList extends React.Component {
         </section>
         <section className='page__section'>
           <div className='heading__wrapper--border'>
-            <h2 className='heading--medium heading--shared-content with-description'>{strings.all_collections} <span className='num--title'>{count ? ` ${tally(count)}` : 0}</span></h2>
+            <h2 className='heading--medium heading--shared-content with-description'>
+              {hasTimeFilter ? strings.active_collections : strings.all_collections}
+              <span className='num--title'>{count ? `${tally(count)}` : 0}</span>
```
chuckwondo, line 103: Now that the leading space is removed, there should be no need for a template string.

PR #733, opened 2020-04-27
CUMULUS-1810: Unify state-change pattern used within reducer functions

chuckwondo:
Unified the coding pattern used for creating Redux reducers to avoid
unnecessary object creation and reduce unnecessary UI component
refreshes.


app/src/js/reducers/api.js
```diff
@@ -1,22 +1,21 @@
...
   REFRESH_TOKEN,
   REFRESH_TOKEN_ERROR,
   REFRESH_TOKEN_INFLIGHT,
   SET_TOKEN
 } from '../actions/types';
-import { createReducer } from '@reduxjs/toolkit';
 
 export const initialState = {
-  authenticated: !!getToken(),
+  authenticated: getToken() != null,
```
flamingbear, line 18: Can you use !== null here?

chuckwondo: I specifically chose `!=` to catch `undefined` as well as `null`, to be safe, but I'm not strongly opposed to changing it to `!==`, if you would prefer it.

flamingbear: I have pretty strong feelings about `==` and `!=` particularly when comparing 0 '' undefined null false and true (see p 121 Javascript the good parts.)

chuckwondo: Yep, fair enough.

flamingbear:
Can you update DEVELOPMENT.md with the new style of reducer.  I know it's super outdated, but if we can keep it a little fresher it might be useful.  particularly `The reducer doesn’t manipulate the current state, but rather returns a new state object that includes the new data ` and the reducer example there `export function reducer (currentState, action.....`

app/src/js/reducers/api.js
```diff
@@ -27,40 +26,38 @@ export const initialState = {
...
+  [LOGOUT]: (draftState) => {
+    draftState.authenticated = false;
   },
-  [REFRESH_TOKEN]: (state, action) => {
-    set(state, 'tokens.error', null);
-    set(state, 'tokens.inflight', false);
-    set(state, 'tokens.token', action.token);
-    setToken(action.token);
+  [REFRESH_TOKEN]: ({ tokens }, { token }) => {
+    setToken(tokens.token = token);
```
flamingbear, line 49: Is this clearer than in two steps?  Because I have to go look up what it means.

flamingbear: I would very much prefer a line that sets draftstate.tokens.token = token, and another that sets the window store's token value.

app/src/js/reducers/api.js
```diff
@@ -27,40 +26,38 @@ export const initialState = {
 };
 
 export default createReducer(initialState, {
-  [DELETE_TOKEN]: (state) => {
-    set(state, 'tokens.token', null);
+  [DELETE_TOKEN]: ({ tokens }) => {
```
flamingbear, line 29: I don't love destructuring the input arguments, but I see you did it everywhere.  Can you tell me why this is clearer or better than using draftState as the input argument?

app/src/js/reducers/api.js
```diff
@@ -27,40 +26,38 @@ export const initialState = {
...
+    setToken(tokens.token = token);
+    tokens.error = null;
+    tokens.inflight = false;
   },
-  [REFRESH_TOKEN_ERROR]: (state, action) => {
-    set(state, 'tokens.error', action.error);
-    set(state, 'tokens.inflight', false);
+  [REFRESH_TOKEN_ERROR]: ({ tokens }, { error }) => {
+    tokens.error = error;
+    tokens.inflight = false;
```
flamingbear, line 55: Would this really be so bad if it was:
```
  [REFRESH_TOKEN_ERROR]: (draftState, action) => {
    draftState.tokens.error = action.error;
    draftState.tokens.inflight = false;
  },
```

app/src/js/reducers/api.js
flamingbear: Would it be so bad as:
 ```
 [DELETE_TOKEN]: ( draftState ) => {
    setToken('');
    draftState.tokens.token = null;
  },
```

File: app/src/js/reducers/api.js (hunk 2 above)
Comment (flamingbear) [inline app/src/js/reducers/api.js:49]:
```suggestion
      setToken(token);
    tokens.token = token;

```

app/src/js/reducers/collections.js
```diff
@@ -1,156 +1,106 @@
...
+  updated: {},
 };
 
 export default createReducer(initialState, {
-  [COLLECTION]: (state, action) => {
-    const { id, data } = action;
+  [COLLECTION]: ({ deleted, map }, { id, data }) => {
     const { name } = deconstructCollectionId(id);
     const collection = data.results.find((element) => element.name === name);
-    set(state, ['map', id, 'inflight'], false);
```
flamingbear, line 51: what happened to this line?
`set(state, ['map', id, 'inflight'], false);`

app/src/js/reducers/collections.js
```diff
@@ -1,156 +1,106 @@
...
-    set(state, ['map', id, 'data'], assignDate(collection));
-    del(state, ['deleted', id]);
-  },
-  [COLLECTION_INFLIGHT]: (state, action) => {
-    const { id } = action;
-    set(state, ['map', id, 'inflight'], true);
-  },
-  [COLLECTION_ERROR]: (state, action) => {
-    const { id } = action;
-    set(state, ['map', id, 'inflight'], false);
```
flamingbear, line 61: Same here?  Were these overlooked or removed on purpose?
`set(state, ['map', id, 'inflight'], false);	`

app/src/js/reducers/collections.js
```diff
@@ -1,156 +1,106 @@
...
-    set(state, ['map', id, 'error'], action.error);
-  },
-
-  [COLLECTION_APPLYWORKFLOW]: (state, action) => {
-    const { id } = action;
-    set(state, ['executed', id, 'status'], 'success');
-    set(state, ['executed', id, 'error'], null);
```
flamingbear, line 68: so I started to look at the createCreateStatusReducer function, and I don't see how it sets the state.executed.id.error here.

I don't think this is working as you expect.
I added this test to the develop branch `test/reducers/collection.js`
It doesn't pass on this branch.

```
test('reducers/collections COLLECTION_APPLYWORKFLOW', (t) => {
  const testState = { ...initialState };
  const action = {
    type: 'COLLECTION_APPLYWORKFLOW',
    id: 'someid'
  };

  const expected = { ...initialState };
  expected.executed.someid = { status: 'success', error: null };

  const actual = reducer(testState, action);

  t.deepEqual(expected, actual);
});
```

flamingbear: I pushed it up on a PR.
https://github.com/nasa/cumulus-dashboard/pull/738
You should make sure it can pass on this branch as well.

app/src/js/reducers/api.js
chuckwondo: Yeah, I'm not convinced I like it either. There was one place where it seemed easier/nicer at the time, so then I used it throughout that file (don't recall which one), and then realized I should do it everywhere for consistency.

However, I think it is perhaps not so good, so I'll get rid of all of the argument destructuring. In fact, at one point I do recall forgetting what the original arguments were supposed to be because everything was destructured, so that also seems to be a strong argument for getting rid of it.

Thanks for calling it out!

chuckwondo: Oh, I'll also get rid of that inline assignment too.

app/src/js/reducers/api.js
chuckwondo: Agreed. Will do.

app/src/js/reducers/collections.js
chuckwondo: I'll go take a look, but I believe the test fails because that nested object becomes
```
{ status: 'success' }
```
instead of
```
{ status: 'success', error: null }
```
There's no need to include `error: null`.

app/src/js/reducers/collections.js
chuckwondo: No, the `inflight` property is _removed_, rather than set to `false`, which achieves the same effect.

app/src/js/reducers/collections.js
chuckwondo: No, the `inflight` property is _removed_, rather than set to `false`, which achieves the same effect.

app/src/js/reducers/collections.js
chuckwondo: I restored this explicit setting of `inflight` to `false`

app/src/js/reducers/collections.js
chuckwondo: I restored this.

app/src/js/reducers/collections.js
chuckwondo: I restored this.

app/src/js/reducers/providers.js
```diff
@@ -54,137 +52,98 @@ export const initialState = {
...
-    const options = data.results.reduce((obj, provider) => {
-      // Several `results` items can share a `providerName`, but
-      // these are de-duplciated by the key-value structure
-      obj[provider.providerName] = provider.providerName;
-      return obj;
-    }, { '': '' });
-    set(state, ['dropdowns', 'group', 'options'], options);
-  },
-  [OPTIONS_PROVIDERGROUP_INFLIGHT]: () => { },
+    state.dropdowns.group = {
```
flamingbear, line 132: again, the group doesn't have any other data besides options on it?

app/src/js/reducers/reconciliation-reports.js
```diff
@@ -1,82 +1,70 @@
...
   createReportInflight: false,
   search: {},
-  deleted: {}
+  deleted: {},
 };
 
 export default createReducer(initialState, {
   [RECONCILIATION]: (state, action) => {
-    const { id, data } = action;
-    set(state, ['map', id, 'inflight'], false);
```
flamingbear, line 37: ^^

flamingbear: I'm pretty sure you're just going to point me to where this is getting set in the code or something. Because all of them are working.

app/src/js/reducers/rules.js
```diff
@@ -1,202 +1,125 @@
...
+    delete state.deleted[action.id];
   },
   [RULE_INFLIGHT]: (state, action) => {
-    const { id } = action;
-    set(state, ['map', id, 'inflight'], true);
+    state.map[action.id] = { inflight: true };
   },
   [RULE_ERROR]: (state, action) => {
-    const { id } = action;
-    set(state, ['map', id, 'inflight'], false);
```
flamingbear, line 76: same same these three

test/reducers/dist.js
```diff
@@ -61,20 +59,13 @@ test('reducers/dist/dist_apigateway', (t) => {
     }
   };
   const newState = reducer(initialState, action);
-  t.deepEqual(expectedState, newState);
```
flamingbear, line 64: I'd prefer to add some junk objects for s3Access and teaLambda just to show they aren't changed by the reducer, but this is fine.

test/reducers/dist.js
```diff
@@ -112,22 +102,15 @@ test('reducers/dist/dist_apilambda', (t) => {
...
   };
-
-  const expectedState = {
-    apiLambda: {
-      inflight: true
-    }
-  };
   const newState = reducer(initialState, action);
-  t.deepEqual(expectedState, newState);
+  t.deepEqual(true, newState.apiLambda.inflight);
```
flamingbear, line 113: why not t.true() like you did above?

test/reducers/dist.js
```diff
@@ -233,29 +204,21 @@ test('reducers/dist/dist_s3access', (t) => {
...
-      inflight: true
-    }
+    s3Access: {}
   };
+  const action = { type: DIST_S3ACCESS_INFLIGHT };
   const newState = reducer(initialState, action);
-  t.deepEqual(expectedState, newState);
+
+  t.false(newState.apiLambda.inflight);
+  t.true(newState.s3Access.inflight);
```
flamingbear, line 221: I think I still prefer the explicit expected state to compare like in the old test, but again. you don't have to change it.

app/src/js/reducers/api.js
```diff
@@ -1,66 +1,64 @@
...
   },
   [REFRESH_TOKEN_INFLIGHT]: (state) => {
-    set(state, 'tokens.inflight', true);
+    state.tokens.inflight = true;
   },
   [SET_TOKEN]: (state, action) => {
-    set(state, 'tokens.token', action.token);
-    setToken(action.token);
-  }
+    setToken((state.tokens.token = action.token));
```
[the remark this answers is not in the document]
chuckwondo: Oops. Not sure what happened there.

app/src/js/reducers/utils/reducer-creators.js
```diff
@@ -0,0 +1,178 @@
...
+ * @example
+ * createReducer(initialState, {
+ *   [ACTION_NAME]: createSerialReducer(
+ *     createSuccessReducer('enabled'),
+ *     createClearItemReducer('disabled')
+ *   ),
+ *   ...
+ * }
+ *
+ * @param  {...any} reducers - list of reducer functions to be invoked in order
```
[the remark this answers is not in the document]
chuckwondo: Right, but I'll add a few more words to the jsdoc to add clarity.

app/src/js/reducers/granules.js
```diff
@@ -64,159 +59,81 @@ export const initialState = {
...
     const options = action.data.results.reduce(
       (obj, { name, version }) =>
         Object.assign(obj, {
-          [`${name} ${version}`]: getCollectionId({ name, version })
+          [`${name} ${version}`]: getCollectionId({ name, version }),
         }),
       {}
     );
-    set(state, ['dropdowns', 'collectionName', 'options'], options);
+    state.dropdowns.collectionName = { options };
```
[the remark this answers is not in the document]
chuckwondo: In this case, the purpose of the options is strictly for holding the dropdown options for a specific collection, so there are no other properties associated it. However, I'll restore the previous behavior, just in case other properties might be added later.

app/src/js/reducers/reconciliation-reports.js
chuckwondo: Yeah, as we chatted about before, nothing is checking _explicitly_ for the value `false`, but rather only for a "falsy" value, which is why simply removing the property works, but I am updating all places where you flagged this to explicitly set to `false` instead of removing the property.

test/reducers/dist.js
chuckwondo: Yep, probably an oversight.

test/reducers/dist.js
```diff
@@ -44,6 +44,7 @@ test('reducers/dist/dist_apigateway', (t) => {
   };
 
   const expectedState = {
+    ...initialState,
```
flamingbear, line 47: This is what I was thinking.  Nice. Thanks.

app/src/js/reducers/granules.js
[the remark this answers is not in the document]
flamingbear: Ah ok, I could have accepted either answer, thanks for knowing it.

app/src/js/reducers/reconciliation-reports.js
flamingbear: ok.  Hope it wasn't too painful for you.

flamingbear:
@chuckwondo All I need to approve is the reducers/api.js L62 fixed and it looks great.

PR #734, opened 2020-04-27
Cumulus 1677 modal workflow updates

flamingbear:
# Track reingested granules on the dashboard.

- Updates the modal workflow for a reingest action on the granule page.
  + When there is a single granule reingested, on complete action the user is redirected to that granule's overview page.
  + When multiple granules are ingested the complete action redirects the user to the granule's running page `/granule/processing`
  
- Adds reingest bulk action to the collections overview page with same behavior as for the granules page.

### DIFF
--- a/app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
+++ b/app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
@@ -72,7 +72,8 @@ export class BatchCommand extends React.Component {
         selected,
         history,
         isOnModalConfirm: true,
-        isOnModalComplete: false
+        isOnModalComplete: false,
+        setState: this.setState.bind(this)
       });
       this.setState({ modalOptions });
 
@@ -137,7 +138,8 @@ export class BatchCommand extends React.Component {
         selected,
         results,
         error,
-        isOnModalComplete: true
+        isOnModalComplete: true,
+        setState: this.setState.bind(this)
       });
       this.setState({ modalOptions });
     }
@@ -154,7 +156,8 @@ export class BatchCommand extends React.Component {
     if (typeof getModalOptions === 'function') {
       const modalOptions = getModalOptions({
         selected,
-        history
+        history,
+        setState: this.setState.bind(this)
       });
       this.setState({ modalOptions });
     }
--- a/app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
+++ b/app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
@@ -43,6 +43,11 @@ export class BatchCommand extends React.Component {
     this.cleanup = this.cleanup.bind(this);
     this.isInflight = this.isInflight.bind(this);
     this.handleClick = this.handleClick.bind(this);
+    this.closeModal = this.closeModal.bind(this);
+  }
+
+  closeModal () {
+    this.setState({ activeModal: false });
   }
 
   componentDidUpdate () {
@@ -73,7 +78,7 @@ export class BatchCommand extends React.Component {
         history,
         isOnModalConfirm: true,
         isOnModalComplete: false,
-        setState: this.setState.bind(this)
+        closeModal: this.closeModal
       });
       this.setState({ modalOptions });
 
@@ -139,7 +144,7 @@ export class BatchCommand extends React.Component {
         results,
         error,
         isOnModalComplete: true,
-        setState: this.setState.bind(this)
+        closeModal: this.closeModal
       });
       this.setState({ modalOptions });
     }
@@ -157,7 +162,7 @@ export class BatchCommand extends React.Component {
       const modalOptions = getModalOptions({
         selected,
         history,
-        setState: this.setState.bind(this)
+        closeModal: this.closeModal
       });
       this.setState({ modalOptions });
     }
--- a/app/src/js/components/Executions/execution-status-graph.js
+++ b/app/src/js/components/Executions/execution-status-graph.js
@@ -93,7 +93,7 @@ class ExecutionStatusGraph extends React.Component {
           className='default-modal execution__modal--visual'>
           <Modal.Header>
             <div className='header' onClick={this.onHide}>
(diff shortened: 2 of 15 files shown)


app/src/js/components/Executions/execution-status-graph.js
```diff
@@ -93,7 +93,7 @@ class ExecutionStatusGraph extends React.Component {
           className='default-modal execution__modal--visual'>
           <Modal.Header>
             <div className='header' onClick={this.onHide}>
-              <div>Click <FontAwesomeIcon icon='compress'/> to return to execution view</div>
+              <div>Click to return to execution view</div>
```
flamingbear, line 96: Missed in an old PR.

jjmccoy:
In Review: UI/UX Review

jjmccoy:
Functionality and workflow looks good. I just have a concern about the display of the bulk reingest past 5 or more granules but it might be something we have to iterate on as we add these bulk modal requests.
Screen Shot 2020-04-28 at 9 15 35 AM

jjmccoy:
UI/UX approved!

app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
```diff
@@ -72,7 +72,8 @@ export class BatchCommand extends React.Component {
         selected,
         history,
         isOnModalConfirm: true,
-        isOnModalComplete: false
+        isOnModalComplete: false,
+        setState: this.setState.bind(this)
```
dopeters, line 76: I don't love the idea of passing `setState` like this. My preference would be to do something like creating this function and passing that to `getModalOptions`:

```
function handleModalClose () {
  this.setState({ activeModal: false });
}
```

Especially since setState controls the `BatchAsyncCommand` component as a whole rather than just the modal.

We can leave this as is, but it worries me a little.

app/src/js/components/ReingestGranules/BatchReingestCompleteContent.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const maxDisplayed = 10;
+
+const BatchReingestCompleteContent = ({
+  results,
+  error
+}) => {
+  const confirmation = `successfully reingested ${results.length > 1 ? 'these' : 'this'} granule${results.length > 1 ? 's' : ''}`;
```
dopeters, line 10: Not necessary, but perhaps we should should just place this directly in the markup so we know results is defined when checking for length

app/src/js/components/ReingestGranules/BatchReingestCompleteContent.js
```diff
@@ -0,0 +1,37 @@
...
+    <>
+      {(results && results.length > 0) &&
+       <>
+         <p>{confirmation}</p>
+         <ul>
+           {results.map((result, index) => {
+             if (index < maxDisplayed) {
+               return <li key={index}>{result}</li>;
+             } return <></>;
+           })}
```
dopeters, line 21: If the results is a lot longer than `maxDisplayed`, do we still want to be iterating over them? Maybe we can do a for loop where the max is the shorter of `results.length` and `maxDisplayed`?

app/src/js/components/ReingestGranules/BatchReingestConfirmContent.js
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <>
+      {requestText}
+      <ul>
+        {selected.map((selection, index) => {
+          if (index < maxDisplayed) {
+            return <li key={index}>{selection}</li>;
+          }
+          return <></>;
+        })}
```
dopeters, line 22: Same note as above re: `maxDisplayed`

flamingbear:
> Functionality and workflow looks good. I just have a concern about the display of the bulk reingest past 5 or more granules but it might be something we have to iterate on as we add these bulk modal requests.
> Screen Shot 2020-04-28 at 9 15 35 AM

Yes, I just picked 10 as a number.  We can definitely make it better however you like, I just didn't want an infinitely long modal and wasn't sure if a scroll was the right way to go.

app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
flamingbear: That's actually fine (and good) and I don't mind fixing it at all. I didn't like it either that much but justified it by thinking we might want to control other parts of the state, but we can do those similarly in the future.

app/src/js/components/ReingestGranules/BatchReingestCompleteContent.js
flamingbear: yeah, good point.  But what if it's was function instead. to keep the markup cleaner?  feelings?

dopeters: Sure!

app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
dopeters: Cool, sounds good!

flamingbear:
@dopeters Think I have addressed your comments. ready for re-review. Thanks.

app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
flamingbear: e53a3b4a

app/src/js/components/ReingestGranules/BatchReingestCompleteContent.js
flamingbear: e53a3b4a

app/src/js/components/ReingestGranules/BatchReingestCompleteContent.js
flamingbear: e53a3b4a

app/src/js/components/ReingestGranules/BatchReingestConfirmContent.js
flamingbear: e53a3b4a

PR #736, opened 2020-04-28
[CUMULUS-1685] Edit rule tests

dopeters:
This ticket was originally for a bug where rule edits were silently ignored/not displaying an error for the user in the UI.

That seemed to be fixed already, so I updated some tests and made other tiny tweaks around the edit flow for various records.


jjmccoy:
LGTG. Error behavior is working. The edit flows look good. Just noticed a typo for 'Continue Editing' for the buttons.

Screen Shot 2020-04-28 at 5 06 16 PM
Screen Shot 2020-04-28 at 5 06 56 PM

PR #737, opened 2020-04-29
CUMULUS-1816: URL Params reflect Datepicker on initial load

bt6453:
So Far:
* Fix string in Datepicker to use constant
* Update Datepicker to add timestamp to url params on initial page load

TODO:
* Is it ok that the timestamp gets changed to "Custom" when navigating to another page?
* Update Datepicker to only be 'Recent' timeframe on home page load

### DIFF
--- a/app/src/js/components/Datepicker/Datepicker.js
+++ b/app/src/js/components/Datepicker/Datepicker.js
@@ -40,7 +40,7 @@ const updateDatepickerStateFromQueryParams = (props) => {
 
     values.dateRange = dropdownValue(values);
     props.dispatch({
-      type: 'DATEPICKER_DATECHANGE',
+      type: DATEPICKER_DATECHANGE,
       data: { ...props.datepicker, ...values }
     });
   }
--- a/app/src/js/components/Datepicker/Datepicker.js
+++ b/app/src/js/components/Datepicker/Datepicker.js
@@ -63,8 +63,7 @@ class Datepicker extends React.PureComponent {
 
   componentDidMount () {
     updateDatepickerStateFromQueryParams(this.props);
-    console.log('PROPS: ' + JSON.stringify(this.props));
-    this.updateQueryParams(this.props);
+    this.homePageInitialDateTime();
   }
 
   refresh (e) {
@@ -77,6 +76,14 @@ class Datepicker extends React.PureComponent {
   clear () {
     const { value, label } = allDateRanges.find((a) => a.label === 'Custom');
     this.props.dispatch(this.dispatchDropdownUpdate(value, label));
+  }
+
+  homePageInitialDateTime () {
+    const { location, queryParams } = this.props;
+    if (location.pathname === '/' && isEmpty(queryParams)) {
+      const { value, label } = allDateRanges.find((a) => a.label === 'Recent');
+      this.props.dispatch(this.dispatchDropdownUpdate(value, label));
+    }
   }
 
   dispatchDropdownUpdate (value, label) {
@@ -279,7 +286,8 @@ Datepicker.propTypes = {
   setQueryParams: PropTypes.func,
   onChange: PropTypes.func,
   dispatch: PropTypes.func,
-  hideWrapper: PropTypes.bool
+  hideWrapper: PropTypes.bool,
+  location: PropTypes.object
 };
 
 export default withRouter(
--- a/app/src/js/components/Datepicker/Datepicker.js
+++ b/app/src/js/components/Datepicker/Datepicker.js
@@ -63,6 +63,8 @@ class Datepicker extends React.PureComponent {
 
   componentDidMount () {
     updateDatepickerStateFromQueryParams(this.props);
+    console.log('PROPS: ' + JSON.stringify(this.props));
+    this.updateQueryParams(this.props);
   }
 
   refresh (e) {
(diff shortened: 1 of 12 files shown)


bt6453:
Updated so that it will only automatically adjust to the "Recent" timeframe on home page load. Every other page will default to "All" if no url parameters are set.

bt6453:
Ah ok. I misunderstood our earlier discussion then. I'll look into making it a one time thing on log in.

flamingbear:
for the questions: 
"Is it ok that the timestamp gets changed to "Custom" when navigating to another page?" - it would be better if it didn't, but doesn't seem like it's worth too much effort if it's not easy to implement. 

"Update Datepicker to only be 'Recent' timeframe on home page load" - Yes, sorry I didnt even see this question. That's what I was talking about I think.

cypress/integration/collections_spec.js
```diff
@@ -38,7 +38,7 @@ describe('Dashboard Collections Page', () => {
 
     it('should display a link to view collections', () => {
       cy.contains('nav li a', 'Collections').as('collections');
-      cy.get('@collections').should('have.attr', 'href', '/collections');
+      cy.get('@collections').should('have.attr', 'href').and('match', /\/collections/);
```
flamingbear, line 41: Is it possible to add `?startDateTime` to this match, just so we know to expect it?  NBD if not.

app/src/js/components/Datepicker/Datepicker.js
```diff
@@ -63,6 +63,7 @@ class Datepicker extends React.PureComponent {
 
   componentDidMount () {
     updateDatepickerStateFromQueryParams(this.props);
+    this.homePageInitialDateTime();
```
flamingbear, line 66: I think this is probably going to get removed.

cypress/integration/main_page_spec.js
```diff
@@ -281,5 +281,13 @@ describe('Dashboard Home Page', () => {
...
     });
+
+    it('adds the default datepicker options to the URL', () => {
+      const now = Date.UTC(2009, 0, 5, 13, 35, 3);
+      cy.clock(now);
+
+      cy.visit('/');
+      cy.url().should('include', 'startDateTime=20090104133500');
+    });
   });
```
flamingbear, line 292: We should also have the test where you visit('/'), "clear all" then click to granules (not "visit"), then click back to home and ensure the url is free of datetime QPs.

cypress/integration/collections_spec.js
flamingbear: Maybe only update one, because they're on all the pages.

test/reducers/datepicker.js
```diff
@@ -16,16 +16,16 @@ test('verify initial state', (t) => {
   t.deepEqual(actual, inputstate);
 });
 
-test('reducer sets initial state to "Recent" on first initialization run.', (t) => {
-  const testStart = Date.now();
-  sinon.useFakeTimers(testStart);
+test('reducer sets initial state to "Custom" on first initialization run.', (t) => {
```
flamingbear, line 19: I don't think this is true anymore.  It should just determine the label from the props I think.(?)

cypress/integration/providers_spec.js
```diff
@@ -30,7 +30,7 @@ describe('Dashboard Providers Page', () => {
 
     it('should display a link to view providers', () => {
```
flamingbear, line 31: Can we add a test in one of the specs where we navigate directly to a "bookmarked url" and then verify that the Datepicker on that page reflects the state of the input URL?

app/src/js/components/Datepicker/Datepicker.js
bt6453: I updated it a little bit, but should it be combined with `updateDatepickerStateFromQueryParams`? Although maybe it makes sense to keep their functions separate

cypress/integration/collections_spec.js
bt6453: With the new one of doing this, I was able to revert back most tests that had to be changed, but updated this for Collections spec

cypress/integration/main_page_spec.js
bt6453: Added

cypress/integration/providers_spec.js
bt6453: Added

test/reducers/datepicker.js
bt6453: Yeah I think I worded this wrong for what I'm trying to test, which is just the initial state. Changed the title and cleaned it up.

app/src/js/components/Datepicker/Datepicker.js
flamingbear: This is fine.

cypress/integration/main_page_spec.js
flamingbear: sorry totally missed that in the last go around.

cypress/integration/main_page_spec.js
```diff
@@ -281,5 +281,48 @@ describe('Dashboard Home Page', () => {
...
+      cy.get('[data-cy=datetime-clear]').click();
+      cy.contains('nav li a', 'Collections').as('collections');
+      cy.get('@collections').click();
+      cy.get('.logo').click();
+
+      cy.url().should('not.include', 'startDateTime=');
+    });
+
+    it('should update the Datepicker with the params in the URL', () => {
+      cy.visit('/?startDateTime=20081229133500&endDateTime=20090105133500');
```
flamingbear, line 309:
```suggestion
      cy.visit('/granules/?startDateTime=20081229133500&endDateTime=20090105133500');
```
This way we're testing an additional page beyond the homepage.  Take it or leave it, but I like it.

bt6453: Sure, that works.

PR #742, opened 2020-04-30
[CUMULUS-1767] Fix lack of error messaging when attempting to update collection name or version

dopeters:
**Description of bug:** When editing the name or version of a collection, there is an error in the console, but it is not displayed in the modal or error message banner

**Solution:** This was happening because the request to `updateCollection` was using the name and version from the payload, regardless of the original collection the user is trying to edit. This causes the request to 404, since that new collection does not exist.

Additionally, in the redux state, that new collection id would be added to `collections.updated` instead of the id of the original collection.  `EditRaw` attempts to grab the error from `collections.updated[ORIGINAL COLLECTION ID]`, but it is undefined, since the request went to the new collection. Instead, `collections.updated[NEW COLLECTION ID]` is defined.

To get the correct error behavior, we need a way for `updateCollection` to know the original collection id. The solution is to make `name` and `version` optional params for that function. That way, if they are present, they will override the values from the payload. Therefore, `collections.updated[ORIGINAL COLLECTION ID]` is no longer undefined, the correct 400 error is returned from the API, and that error is displayed.

https://nasa.github.io/cumulus-api/#updatereplace-collection

Change ended up being relatively simple. Also added a test to verify the error is properly displayed.

### DIFF
--- a/app/src/js/actions/index.js
+++ b/app/src/js/actions/index.js
@@ -158,12 +158,13 @@ export const createCollection = (payload) => ({
   }
 });
 
-export const updateCollection = (payload) => ({
+// include the option to specify the name and version of the collection to update in case they differ in the payload
+export const updateCollection = (payload, name, version) => ({
   [CALL_API]: {
     type: types.UPDATE_COLLECTION,
     method: 'PUT',
-    id: getCollectionId(payload),
-    path: `collections/${payload.name}/${payload.version}`,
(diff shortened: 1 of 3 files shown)


jjmccoy:
LGTG for Collection Edit Error messaging when changing a name or version. Validation worked fine.
Screen Shot 2020-05-01 at 12 11 48 PM

PR #743, opened 2020-04-30
Cumulus 1816 behavior

flamingbear:
Only for example behavior.  You might know a much better way to do this.  But this was what hit my head first.


bt6453:
This is good, but it still fills in the query params if the home page is refreshed. I think I'm going to work off this and see what I can do.

flamingbear:
> This is good, but it still fills in the query params if the home page is refreshed. I think I'm going to work off this and see what I can do.

when refreshed with a F5 full browser refresh or something else?

bt6453:
Yeah with full refresh from the refresh button or cmd+R

flamingbear:
@bt6453 your solution was much better.  Closing this

PR #746, opened 2020-05-05
CUMULUS-1777: Update dashboard to log out user on 401/403 response from API

markdboyd:


app/src/js/components/Header/header.js
```diff
@@ -33,8 +33,10 @@ class Header extends React.Component {
 
   componentDidMount () {
     const { dispatch, api } = this.props;
-    if (api.authenticated) dispatch(getApiVersion());
-    dispatch(getCumulusInstanceMetadata());
+    if (api.authenticated) {
+      dispatch(getApiVersion());
+      dispatch(getCumulusInstanceMetadata());
```
markdboyd, line 38: This is unrelated to the rest of the changes. However, I kept getting an error message of "Authorization header missing" on my homepage when initially loading the dashboard. 

I tracked it down to this header component making the request for instance metadata, even when the user is not logged in. Since the /instanceMeta endpoint does require authentication, this change seemed reasonable and made the error go away for me. 

I can exclude this change if there's some reason it shouldn't be made though.

app/src/js/middleware/request.js
```diff
@@ -10,30 +7,28 @@ import {
...
+  requestAction,
+  statusCode
+}, next) => {
   console.groupCollapsed('handleError');
   console.log(`id: ${id}`);
   console.log(`type: ${type}`);
   console.dir(error);
   console.dir(requestAction);
   console.groupEnd();
-  if (error.message) {
```
markdboyd, line 20: All of this code is major technical debt and as far as I can tell was untested anywhere. So I ripped it out in favor of the change to log out the user when we receive specific error response codes from the API

test/middleware/request.js
```diff
@@ -1,98 +1,118 @@
...
+      token
     }
-  });
-  t.context.nextHandler = requestMiddleware({dispatch: doDispatch, getState: doGetState});
+  }
+}));
+const dispatchStub = sinon.stub();
+const nextStub = sinon.stub();
+
+const create = () => {
```
markdboyd, line 28: Since I was updating the request middleware, I needed to add tests. While I was in here, it made sense to update the testing pattern to match the Redux docs:

https://redux.js.org/recipes/writing-tests#middleware

I also think the tests are a lot more readable than they were before, but I guess it's subjective

test/middleware/request.js
```diff
@@ -135,15 +167,51 @@ test.cb('should dispatch error action for failed request', (t) => {
...
 
-  const actionHandler = t.context.nextHandler(action => {
-    t.deepEqual(action, expectedAction);
-    t.end();
-  });
+  await invokeMiddleware(actionObj);
+  t.deepEqual(next.firstCall.args[0], expectedAction);
+});
+
+test.serial('should dispatch login error action for 401 response', async (t) => {
```
markdboyd, line 174: This test and the one below it are the new tests

test/middleware/request.js
```diff
@@ -1,98 +1,118 @@
...
+test.afterEach.always(() => {
+  nock.cleanAll();
+  dispatchStub.resetHistory();
+  getStateStub.resetHistory();
+  nextStub.resetHistory();
+  loginErrorStub.resetHistory();
+});
+
+test.serial('should pass action to next if not an API request action', (t) => {
   const actionObj = {};
```
flamingbear, line 62:
```suggestion
const actionObj = { type: 'ANYTHING', some: 'action' };```
Can we add something just to make sure we're not just dumping out an empty object? (I know we aren't)

File: test/middleware/request.js [hunk 6]
```diff
@@ -102,19 +122,31 @@ test.cb('should be able to use provided authorization headers', (t) => {
...
+    const nextAction = requestPromiseStub.firstCall.args[0];
+    t.deepEqual(nextAction.headers.Authorization, 'Bearer another-token');
+  } finally {
+    revertRequestStub();
+  }
 });
 
-test.cb('should dispatch error action for failed request', (t) => {
+test.serial('should dispatch error action for failed request', async (t) => {
   nock(`http://localhost:${port}`)
```
flamingbear, line 144: I don't know why that was all PORT'd up, but you can make all of the tests use `http://anyhost`?

test/middleware/request.js
flamingbear: This is nice! 👍

flamingbear:
Added a PR with my suggestions https://github.com/nasa/cumulus-dashboard/pull/749/files

test/middleware/request.js
flamingbear: don't commit that suggestion, my comment was included somehow.

PR #753, opened 2020-05-08
[CUMULUS-1948] Add provider to Granule tables

bt6453:


bt6453:
@laurenfrederick Yup, I updated the Granule cypress test to account for the providers column

PR #754, opened 2020-05-08
[CUMULUS-1903] Tabs for reconciliation report buckets 

dopeters:
This adds a TableCards component to the individual reconciliation report view

For now, they are displaying the same tables that were originally displayed on the page. When the API is updated, they should be updated to show the individual buckets.

Also just for naming conventions, changed `sortIdx` to `sortId` in sortable table since it's using a string instead of an index.

Added a simple pagination implementation for tables where the data is not coming from it's own API request. (i.e. for tables that don't need to use the `List` component)

Finally, I removed the interval timers from the reconciliation report as they were causing issues. The built in timer component on the list page is still there. I figured we can revisit this as part of the whole timer solution and the ticket specific for these pages https://bugs.earthdata.nasa.gov/browse/CUMULUS-1877


jjmccoy:
UI/UX: 

- Tabs: Nice implementation of the tab and table interactivity. 

- Pagination: Looks nice. Noticed that when you get to the final page and click on "previous" it doesn't respond and go back in sequence -- it just stays on that page. 

- Timers: Agree on the timers decision.

dopeters:
> UI/UX:
> 
> * Tabs: Nice implementation of the tab and table interactivity.
> * Pagination: Looks nice. Noticed that when you click on "previous" it doesn't respond and go back in sequence -- it just stays on that page.
> * Timers: Agree on the timers decision.

Fixed the issue with the previous page button!

jjmccoy:
UI/UX: Thanks for that quick fix. LGTG.

app/src/js/components/ReconciliationReports/reconciliation-report.js
```diff
@@ -131,9 +110,15 @@ class ReconciliationReport extends React.Component {
...
+    e.preventDefault();
+    this.setState({ activeIdx: index });
+  }
+
   render () {
     const { reconciliationReports } = this.props;
     const { reconciliationReportName } = this.props.match.params;
+    const { activeIdx } = this.state;
 
     const record = reconciliationReports.map[reconciliationReportName];
```
flamingbear, line 123: Nothing to be done here, but I'm noting what an unfortunate choice of name "map"  we did it all over the place.

app/src/js/components/ReconciliationReports/reconciliation-report.js
```diff
@@ -49,25 +36,17 @@ const parseFileObject = (d) => {
...
-  }
-
-  reload (immediate) {
-    const { reconciliationReportName } = this.props.match.params;
-    const { dispatch } = this.props;
-    if (this.cancelInterval) { this.cancelInterval(); }
-    this.cancelInterval = interval(() => dispatch(getReconciliationReport(reconciliationReportName)), updateInterval, immediate);
+    const { dispatch, match } = this.props;
+    const { reconciliationReportName } = match.params;
+    dispatch(getReconciliationReport(reconciliationReportName));
```
flamingbear, line 49: I also know this predates this PR, but why do we dispatch the report every time we display it?  we're also storing it in map and fetching it every time it seems...

app/src/js/components/SortableTable/SortableTable.js
```diff
@@ -144,7 +159,7 @@ const SortableTable = ({
                   {headerGroup.headers.map(column => {
                     return (
                       <div {...column.getHeaderProps()} className='th'>
-                        <div {...column.getSortByToggleProps()} className={`${column.canSort ? 'table__sort' : ''}`}>
+                        <div {...column.getSortByToggleProps()} className={`${column.canSort ? 'table__sort' : ''}${column.isSortedDesc === true ? '--desc' : (column.isSortedDesc === false ? '--asc' : '')}`}>
```
flamingbear, line 162: It's not possible for column to look like `{canSort: false, isSortedDesc: true, ...}`? I would guess not.

app/src/css/vendor/bootstrap/overrides/_card.scss
```diff
@@ -0,0 +1,28 @@
+.card-wrapper {
+  display: flex;
+  justify-content: space-between;
```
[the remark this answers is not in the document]
jjmccoy: This ticket is focused on the implementation and function. We can look at the flex grid and how wide the columns and padding will be in the style ticket.

app/src/js/components/Pagination/pagination.js
```diff
@@ -9,7 +9,6 @@ const disabled = ' pagination__link--disabled';
 class Pagination extends React.Component {
   constructor () {
     super();
-    this.displayName = 'Pagination';
```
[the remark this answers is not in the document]
dopeters: My understanding of displayName is that we only need to explicitly set it if we want to see a specific/different name for debugging https://reactjs.org/docs/react-component.html#displayname

In this case, it matches the name of the component so it seems unnecessary.

app/src/css/vendor/bootstrap/overrides/_card.scss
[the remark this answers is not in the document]
jjmccoy: The style ticket is [1907](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1907).

app/src/js/components/ReconciliationReports/index.js
```diff
@@ -4,27 +4,19 @@ import PropTypes from 'prop-types';
...
 import ReconciliationReportList from './list';
 import ReconciliationReport from './reconciliation-report';
 import withQueryParams from 'react-router-query-params';
 
-const { updateInterval } = _config;
-
 class ReconciliationReports extends React.Component {
   constructor () {
     super();
-    this.displayName = 'Reconciliation Reports';
```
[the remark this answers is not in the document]
dopeters: Same note as above, though if we want the space in there for debugging, I could add this back (in a way that doesn't trigger a DeepScan error)

But yeah, seems like they were just added to all components regardless 🤷

app/src/css/vendor/bootstrap/overrides/_card.scss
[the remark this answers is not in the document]
dopeters: Thanks @jjmccoy! Also, since the number of cards will change once we have the actual bucket names, etc, I'm sure the style implementation will change as well

app/src/js/components/SortableTable/SortableTable.js
dopeters: Ah interesting point. Correct, that would not be possible, though it might make sense to more clearly encapsulate that logic here

Not sure if this is more or less confusing 😂 

```
className={`${column.canSort ? (`table-sort${column.isSortDesc === true ? '--desc' : (column.isSortedDesc === false ? '--asc' : '')}`) : ''}`}
```

or could do something like this

```
let columnClassName = '';
if (column.canSort) {
  columnClassName = `table-sort${column.isSortedDesc === true ? '--desc' : (column.isSortedDesc === false ? '--asc' : '')}`;
}
```

app/src/js/components/ReconciliationReports/reconciliation-report.js
dopeters: My guess would be to ensure we get the freshest version of of the report every time we get to the page. Originally had this wrapped in an if statement so that it would dispatch only if the report was not in the map, but I thought that might have been causing the issue Lauren was seeing. Can add it back though

app/src/js/components/ReconciliationReports/reconciliation-report.js
dopeters: Agreed

app/src/js/components/Pagination/pagination.js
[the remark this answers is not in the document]
flamingbear: thanks for googling that for me. :)

app/src/js/components/ReconciliationReports/reconciliation-report.js
flamingbear: So I was unfamiliar with this, I don't know if the reports change once they're available.  If not we should add it back, if so, this is fine.  let's ask Lauren. @laurenfrederick

app/src/js/components/SortableTable/SortableTable.js
flamingbear: It's fine as it, but I like the last one best. your call.

app/src/css/vendor/bootstrap/overrides/_card.scss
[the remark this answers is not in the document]
flamingbear: gotcha. thanks

app/src/js/components/ReconciliationReports/index.js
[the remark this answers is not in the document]
dopeters: Decided to add it back just so it'll be easier to read when debugging

app/src/js/components/SortableTable/SortableTable.js
dopeters: Yep, agreed, changed it to the last one

app/src/js/components/ReconciliationReports/reconciliation-report.js
dopeters: After discussing with Lauren on scrum, these will never change, so I put this back in the conditional

PR #755, opened 2020-05-08
Fixes unsafe reducer-creators.

flamingbear:
Fixes CUMULUS-1988

Updates the reducer-creators to use safe methods not assuming that all values exist.

Adds tests for a small number of reducers as well as for the reducer-creators.

### DIFF
--- a/app/src/js/reducers/providers.js
+++ b/app/src/js/reducers/providers.js
@@ -91,17 +91,11 @@ export default createReducer(initialState, {
   [UPDATE_PROVIDER]: (state, action) => {
     const { id, data } = action;
     state.map[id] = { data };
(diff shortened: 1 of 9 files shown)


app/src/js/reducers/providers.js
```diff
@@ -102,12 +102,6 @@ export default createReducer(initialState, {
     state.list.inflight = false;
     state.list.error = false;
   },
-  [PROVIDERS]: (state, action) => {
```
laurenfrederick, line 105: Just because I don't know things... why is this removed?

flamingbear: because it was duplicated in the code.

PR #762, opened 2020-05-14
Mhs/cumulus 1869/pre work cleanup

flamingbear:
Sorry this is a combo of both prettier.io the Collection files *and* cleaning up the connect() calls 

I think all of the existing mapStateToProps that were `(state) => state` can cause undue rendering, so I removed unused parts of the state where appropriate.

eslintrc was updated because prettier forces no space between function and paren.  I set it all to ignore for now so that this didn't touch every file.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,7 @@
   },
   "rules": {
     "comma-dangle": ["error", "only-multiline"],
+    "space-before-function-paren": ["warn", {"anonymous": "ignore", "named": "ignore", "asyncArrow": "ignore"}],
     "semi": [2, "always"],
     "no-extra-semi": 2,
     "semi-spacing": [2, { "before": false, "after": true }],
--- a/app/src/js/components/Collections/add.js
+++ b/app/src/js/components/Collections/add.js
@@ -32,7 +32,9 @@ const AddCollection = ({ location = {}, collections, dispatch, schema }) => {
   }, [collectionSchema, collectionId, collectionsMap, isCopy]);
 
   const getBaseRoute = (pk = collectionId) => {
-    return (pk && pk !== 'unknown') ? collectionHref(pk) : '/collections';
+    return pk && pk !== 'unknown'
+      ? collectionHref(pk)
+      : '/collections';
   };
 
   return (
@@ -56,7 +58,9 @@ AddCollection.propTypes = {
   schema: PropTypes.object,
 };
 
-export default withRouter(connect(state => ({
-  collections: state.collections,
-  schema: state.schema
-}))(AddCollection));
+export default withRouter(
+  connect((state) => ({
+    collections: state.collections,
+    schema: state.schema,
+  }))(AddCollection)
+);
--- a/app/src/js/components/Collections/add.js
+++ b/app/src/js/components/Collections/add.js
@@ -32,9 +32,7 @@ const AddCollection = ({ location = {}, collections, dispatch, schema }) => {
   }, [collectionSchema, collectionId, collectionsMap, isCopy]);
 
   const getBaseRoute = (pk = collectionId) => {
-    return pk && pk !== 'unknown'
-      ? collectionHref(pk)
-      : '/collections';
+    return pk && pk !== 'unknown' ? collectionHref(pk) : '/collections';
   };
 
   return (
--- a/app/src/js/components/Collections/edit.js
+++ b/app/src/js/components/Collections/edit.js
@@ -6,7 +6,7 @@ import { withRouter } from 'react-router-dom';
 import {
   getCollection,
   updateCollection,
-  clearUpdateCollection
+  clearUpdateCollection,
 } from '../../actions';
 import { getCollectionId } from '../../utils/format';
 import EditRaw from '../EditRaw/edit-raw';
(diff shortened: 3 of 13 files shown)


app/src/js/components/Collections/add.js
```diff
@@ -32,7 +32,7 @@ const AddCollection = ({ location = {}, collections, dispatch, schema }) => {
   }, [collectionSchema, collectionId, collectionsMap, isCopy]);
```
flamingbear, line 32: format only in this file.

app/src/js/components/Collections/edit.js
```diff
@@ -6,25 +6,27 @@ import { withRouter } from 'react-router-dom';
 import {
```
flamingbear, line 6: format only in this file.

app/src/js/components/Collections/granules.js
```diff
@@ -194,14 +194,13 @@ CollectionGranules.propTypes = {
   granules: PropTypes.object,
```
flamingbear, line 194: formatting above changes below.

app/src/js/components/Collections/index.js
```diff
@@ -17,42 +16,91 @@ import DatePickerHeader from '../../components/DatePickerHeader/DatePickerHeader
...
     return (
-      <div className='page__collections'>
-        <DatePickerHeader onChange={this.query} heading={strings.collections}/>
-        <div className='page__content'>
-          <div className='wrapper__sidebar'>
-            <Route path='/collections/all' component={Sidebar} />
-            <Route path='/collections/edit/:name/:version' component={Sidebar} />
-            <Route path='/collections/collection/:name/:version' component={Sidebar} />
-            <div className={existingCollection ? 'page__content--shortened' : 'page__content'}>
+      <div className="page__collections">
```
flamingbear, line 28: this diff is all format.  the rest of the file is mixed.  I pulled out withQueryParams as it wasn't used.

app/src/js/components/Collections/ingest.js
```diff
@@ -69,59 +68,83 @@ class CollectionIngest extends React.Component {
...
         </section>
       </div>
     );
   }
 
-  renderJson (data) {
+  renderJson(data) {
     return (
       <ul>
         <li>
```
flamingbear, line 150: non-format changes below.

app/src/js/components/Collections/list.js
```diff
@@ -148,10 +154,17 @@ CollectionList.propTypes = {
   mmtLinks: PropTypes.object,
   dispatch: PropTypes.func,
   config: PropTypes.object,
-  datepicker: PropTypes.object
+  datepicker: PropTypes.object,
 };
 
 CollectionList.displayName = 'CollectionList';
```
flamingbear, line 161: format above, changes below

app/src/js/components/Collections/logs.js
```diff
@@ -8,24 +8,35 @@ import LogViewer from '../Logs/viewer';
 import { strings } from '../locale';
```
flamingbear, line 8: format only.

app/src/js/components/Collections/overview.js
```diff
@@ -11,15 +11,15 @@ import {
   getCollection,
```
flamingbear, line 11: formating only

app/src/js/components/locale.js
```diff
@@ -41,6 +41,7 @@ export const strings = new LocalizedStrings({
     view_granules_overview: 'View Granule Overview'
   },
   gitc: {
+    active_collections: 'Active Layers',
```
flamingbear, line 44: random locale string I found was missing.

.eslintrc
```diff
@@ -17,6 +17,7 @@
   },
   "rules": {
     "comma-dangle": ["error", "only-multiline"],
+    "space-before-function-paren": ["warn", {"anonymous": "ignore", "named": "ignore", "asyncArrow": "ignore"}],
```
flamingbear, line 20: this should be until we go all in or not and change them all.

app/src/js/components/Collections/index.js
```diff
@@ -62,11 +110,12 @@ class Collections extends React.Component {
...
+
 Collections.propTypes = {
   children: PropTypes.object,
   dispatch: PropTypes.func,
   location: PropTypes.object,
-  queryParams: PropTypes.object
 };
 
-export default withRouter(withQueryParams()(connect(state => state)(Collections)));
+export default withRouter(connect(null, null)(Collections));
```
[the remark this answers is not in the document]
flamingbear: I think by putting in null, we are getting the default mapDispatchToProps.  (Or that's what I thought was happening) I'll go check.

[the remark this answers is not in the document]
flamingbear: Hmm, Maybe I should call it with no params, I thought was being more specific, but the docs say:

`Your component will receive dispatch by default, i.e., when you do not supply a second parameter to connect():`

But we need dispatch in the component until we remove the `query()`.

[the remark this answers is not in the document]
dopeters: Ah right, we do still need dispatch. Seems like we can call it with no params then, but since either way works, I'll leave it up to you

PR #763, opened 2020-05-14
[CUMULUS-1888] Granule Pagination improvements

bt6453:
So Far:

* Change the Granule CSV Download. Before it would constantly refresh the CSV data in the background, which could take 1.5s with ~5000 granules. Now the data isn't updated until the user clicks the button.


bt6453:
Note: I whitelisted a vulnerable with `http-proxy` (used by `webpack-dev-server`) to get CI to run, but that should be removed before merging in

jjmccoy:
UI/UX: The on click action works. LGTG.

bt6453:
@laurenfrederick Changelog has been updated. Also I looked in the workflows being updated in the background, but couldn't find an easy fix.

PR #765, opened 2020-05-15
[CUMULUS-1990] Display error for rerun rule

dopeters:
The crux of the issue was that we were not pulling errors from rules.rerun in the state so they were not added to the error report. Reenabled the test for this behavior.

Additional cleanup:
1. Imports of `DropdownAsync` were called `AsyncCommands` which was confusing since we have multiple components with that name, so updated all instances of that.
2. Improved success and error callback handling for AsyncCommand. 
3. Removed success timeout so that it would not be called when a success modal should be visible. If there is a modal, it will be called on the close/confirm button click. Otherwise, if there is no modal, `success` is called immediately.


app/src/js/components/Rules/rule.js
```diff
@@ -117,7 +117,8 @@ class Rule extends React.Component {
       get(rules.map, [ruleName, 'error']),
       get(rules.deleted, [ruleName, 'error']),
       get(rules.enabled, [ruleName, 'error']),
-      get(rules.disabled, [ruleName, 'error'])
+      get(rules.disabled, [ruleName, 'error']),
+      get(rules.rerun, [ruleName, 'error'])
```
dopeters, line 121: This is the crux of the fix. Everything else is essentially cleanup

app/src/js/components/AsyncCommands/AsyncCommands.js
```diff
@@ -104,33 +105,32 @@ class AsyncCommand extends React.Component {
...
+                </div>
+              )}
+              <h4>{confirmText}</h4>
+            </div>
+          </DefaultModal>
           <DefaultModal
-            className='link--no-underline'
-            onCancel={this.cancel}
+            className='async__modal--success'
+            onCancel={success}
```
flamingbear, line 127: What does success mean now? It didn't make sense to me that this is connected to `onCancel` now

dopeters: Yeah this is confusing. The modal has two buttons, cancel and confirm. You can use one or the other. Changed this to use the confirm button so that we're using `onConfirm` instead of `onCancel` for clarity.

PR #768, opened 2020-05-18
Mhs/cumulus 1869/fix granule metrics tiles

flamingbear:
Connects datepicker state to the timer the timer and now will the timer will refresh on datepicker changes.

Adds datepicker params to the getCollection call to Cumulus API so that the correct statistics show up when a page is filtered by the datepicker.

Adds cypress `data-cy` selector to overview for testing.

after this PR all of these items will have been addressed.
Fixes: https://bugs.earthdata.nasa.gov/browse/CUMULUS-1869

### DIFF
--- a/app/src/js/components/Collections/overview.js
+++ b/app/src/js/components/Collections/overview.js
@@ -1,8 +1,9 @@
 'use strict';
+import { get } from 'object-path';
+import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
-import { withRouter, Link } from 'react-router-dom';
-import PropTypes from 'prop-types';
+import { Link, withRouter } from 'react-router-dom';
 import {
   clearGranulesFilter,
   clearGranulesSearch,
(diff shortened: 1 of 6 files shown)


app/src/js/components/Collections/overview.js
```diff
@@ -13,30 +14,30 @@ import {
...
 import Breadcrumbs from '../Breadcrumbs/Breadcrumbs';
+import DeleteCollection from '../DeleteCollection/DeleteCollection';
+import Dropdown from '../DropDown/dropdown';
+import SimpleDropdown from '../DropDown/simple-dropdown';
+import Bulk from '../Granules/bulk';
 import ListFilters from '../ListActions/ListFilters';
+import { strings } from '../locale';
+import Overview from '../Overview/overview';
+import Search from '../Search/search';
+import List from '../Table/Table';
```
flamingbear, line 40: These is automatic reordering of the imports, plus adding `isEqual` from lodash.

PR #771, opened 2020-05-20
Mhs/cumulus 1916/seed reconcilation reports

flamingbear:
**Summary:** Updates localAPI database seeding to work with (presumed) cumulus v1.22.0 release that includes recent updates to the reconciliation reports. 

Addresses [CUMULUS-1916](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1916)

## Changes

* moved and renamed sample reconciliation reports from `@cumulus/api/app/data/reconciliation-reports` to cypress/fixtures/seeds/reconciliation-reports
* adds a reconciliation report record fixture
* uploads s3 reconciliation report files to correct locations.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/cypress/integration/auth_spec.js
+++ b/cypress/integration/auth_spec.js
@@ -33,7 +33,7 @@ describe('Dashboard authentication', () => {
 
   it('should logout user on invalid JWT token', () => {
     cy.window().its('appStore').then((store) => {
-      cy.task('generateJWT', {}).then((invalidJwt) => {
+      cy.task('generateJWT', { expirationTime: NaN }).then((invalidJwt) => {
         // Dispatch an action to set the token
         store.dispatch({
           type: SET_TOKEN,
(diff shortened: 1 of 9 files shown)


cypress/integration/auth_spec.js
```diff
@@ -62,7 +62,7 @@ describe('Dashboard authentication', () => {
     });
 
     cy.window().its('appStore').then((store) => {
-      const expirationTime = (new Date(Date.now() - 24 * 3600 * 1000)).valueOf();
+      const expirationTime = (new Date(Date.now() - 24 * 3600 * 1000)).valueOf() / 1000.0;
```
dopeters, line 65: Just curious why we're dividing by 1000.0 here

flamingbear: the old api/lib/token.js took the input expirationTime and converted it to seconds from ms in the code https://github.com/nasa/cumulus/blob/v1.22.1/packages/api/lib/token.js#L5
The latest doesn't do this anymore.
https://github.com/nasa/cumulus/blob/master/packages/api/lib/token.js#L3-L5

PR #772, opened 2020-05-21
CUMULUS-1913: Add datepicker to Reconciliation Reports

jennyhliu:

Addresses [CUMULUS-1913: Add datepicker to Reconciliation Reports](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1913)

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/js/components/locale.js
+++ b/app/src/js/components/locale.js
@@ -34,6 +34,7 @@ export const strings = new LocalizedStrings({
     logo: 'cumulus-logo.png',
     operations: 'Operations',
     pdrs: 'Pdrs',
+    reconciliation_reports: 'Reconciliation Reports',
     remove_from_cmr: 'Remove from CMR',
     running_granules: 'Running Granules',
     total_granules: 'Total Granules',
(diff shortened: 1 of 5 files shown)


app/src/js/components/locale.js
```diff
@@ -34,6 +34,7 @@ export const strings = new LocalizedStrings({
     logo: 'cumulus-logo.png',
     operations: 'Operations',
     pdrs: 'Pdrs',
+    reconciliation_reports: 'Reconciliation Reports',
```
flamingbear, line 37: If we add it here do we need to have it in the gitc block as well?  I think so, but I'm not 100% certain.

flamingbear, line 37: If we add it here do we need to have it in the gitc block as well? I think so, but I'm not 100% certain.

jennyhliu: Thanks, updated gitc block

flamingbear: lgtm

jjmccoy:
UI/UX: The width of the heading is too large with the DTP so the double line is fine. Everything else looks and works fine. LGTG!

PR #775, opened 2020-05-28
[CUMULUS-1915] Add filters for `Report Type` and `Status` to reconcilation-reports page

jennyhliu:

Addresses [CUMULUS-1915: Reconciliation Reports - List - Add filters for "Report Type" and "Status"](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1915)

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


app/src/js/components/ReconciliationReports/list.js
```diff
@@ -106,6 +133,6 @@ ReconciliationReportList.propTypes = {
   reconciliationReports: PropTypes.object
 };
 
-export default withRouter(withQueryParams()(connect(state => ({
```
flamingbear, line 109: 🙌

cypress/fixtures/seeds/reconciliationReportFixture.json
```diff
@@ -20,6 +20,13 @@
       "location": "s3://localbucket/localrun/reconciliation-reports/inventoryReport-20200114T202529026.json",
       "type": "Inventory",
       "status": "Generated"
+    },
+    {
+      "createdAt": 1589898791935,
+      "name": "inventoryReport-20200519T143311876.json",
```
flamingbear, line 26:
```suggestion
      "name": "inventoryReport-20200519T143311876",
```

app/src/js/utils/status.js
```diff
@@ -12,3 +12,9 @@ export const pdrStatus = {
   Completed: 'completed',
   Failed: 'failed'
 };
+
+export const reconciliationReportStatus = {
+  Generated: 'Generated',
+  Pending: 'Pending',
+  Failed: 'Failed'
```
flamingbear, line 19: I think lauren wanted us to use 'conflict' instead of failed. unless failed is the inability to run the report?  So 'conflict' is we generated the report and files don't match, and 'failed' is we somehow were unable to generate the report?

jjmccoy: UI/UX input: This is regarding the status of the report as it is created like we do when a granule is ingested. We need to validate if the report was generated or not. What you are referring to is regarding an individual report after it's generated and the status is the file comparison and missing files. I think we should change the term on the individual report itself from "status" to "state" or something else for less confusion. I'll think on that and propose to OXWG to see what they say.

flamingbear: I like state better. 👍

cypress/fixtures/seeds/reconciliationReportFixture.json
```diff
@@ -20,6 +20,13 @@
...
+      "location": "s3://localbucket/localrun/reconciliation-reports/inventoryReport-20200519T143311876.json",
```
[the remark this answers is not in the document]
jennyhliu: I copied one from my cumulus stack.  The file has initial status 'RUNNING', and the file is updated with the report information later, if anything happens (such as CMR password doens't work), then the report never finishes, stays in 'RUNNING'.

app/src/js/utils/status.js
jennyhliu: Yeah, it's somehow confusion.  I like 'state' also.

cypress/fixtures/seeds/reconciliationReportFixture.json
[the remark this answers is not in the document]
flamingbear: Ok, so that means this is a 'failed' state test, not a conflict.  That file is fine (but you could pretty-print it just because)

app/src/js/utils/status.js
flamingbear: We should definitely stick to correct words here:  status => (passed/conflict), state => (generated/pending/failed).

Although the tenses agree better with,  pass/conflict, I still like passed/conflict better. but what do I know.

flamingbear: @jjmccoy will you get back to us with the OXWG decision on that? We can write a ticket to ammend.  I will approve the rest of the current changes.

flamingbear:
thanks for clarifying that for me @jennyhliu !

app/src/js/utils/status.js
jjmccoy: Okay will do.

PR #776, opened 2020-05-28
Removes datetimepicker's initial value

flamingbear:
**Summary:** Stops the datetimepicker from setting the initial value to "recent" data when loading the application the first time.

Addresses [CUMULUS-1994: default datetime filters should be none.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1994)

## Changes

* removed homePageInitialDateTime()
* removed any reference to `new_session`
* updated tests

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


cypress/support/commands.js
```diff
@@ -115,3 +115,10 @@ Cypress.Commands.add('expectDeepEqualButNewer', (inewObject, ifixtureObject) =>
 Cypress.Commands.add('clearStartDateTime', () => {
   cy.get('li[data-cy="startDateTime"]').find('.react-datetime-picker__clear-button').click();
 });
+
+/**
+ * Add custom command to set the dropdown value to any value.
+ */
+Cypress.Commands.add('setDropdown', (targetValue) => {
+  cy.get('[data-cy=datetime-dropdown]').select(targetValue);
+});
```
dopeters, line 124: small nitpick, but might be good to name this in a way that makes it clear that this is for the datetime dropdown, rather than something that could be used to set any dropdown

flamingbear: > Looks great, small comment about naming.
> 
> Though I am noticing one thing that we might want to look into later. If I go to the granules page and select "Recent" from the dropdown, everything is fine. If I then click to the collections page, it still works fine, but the dropdown now says "Custom". Even if it was clicked right away. I think it has do with Recent only being applicable when the time is exactly a week from the current time, so nothing is actually broken, but it is a little bit of a confusing user experience.

I had thought that I had solved that previously. I will look into it, it seems like we should be able to determine if it's still '1 week' exactly.

jjmccoy:
UI/UX: LGTG

cypress/support/commands.js
flamingbear: 182e03f4

PR #779, opened 2020-05-29
CUMULUS-1977: bulk granule modals

markdboyd:
Corresponding UI changes for https://github.com/nasa/cumulus/pull/1668

This PR updates the "bulk granules" modal to allow you to choose between the two types of bulk granule operations:

- Bulk granule operations (running workflows on granules)
- Bulk granule deletion (deleting granules)

Depending on your choice, a modal opens to allow you to generate and submit that type of request.

Misc fixes:

- The error state from a previous failed request should get removed when you close the modal for a bulk operation/delete request. My thinking here was that once you've closed the modal and re-opened it, you don't need to see the failure from your previous request
- The error state from bad JSON in the form gets cleaned up as edit it
- ~The selected granules from the overview page are passed through as the value for `ids` in the JSON for bulk granules/operations requests~
  - I couldn't get this feature to work reliably without having it overwrite what you're typing in the query box as you type. Since it is a nice to have and not a requirement of the ticket, I removed it

Tests:

- [x] Reducer tests for new action types
- [x] Component tests for new pure functional components
- [x] Integration tests


app/src/css/_buttons.scss
```diff
@@ -1,3 +1,7 @@
+$delete-btn-icon: '\f187';
```
markdboyd, line 1: I didn't actually change much here but my edtior did some auto-formatting on save. I just moved some definition for delete button styles into variables for easier re-use in the bulk granule delete modal.

app/src/js/components/Granules/bulk-granule-delete.js
```diff
@@ -0,0 +1,138 @@
...
+  requestId,
+  showModal,
+  selected = [],
+  success
+}) => {
+  const [query, setQuery] = useState(JSON.stringify(defaultQuery, null, 2));
+  const [errorState, setErrorState] = useState();
+
+  // TODO: is there a better way to do this?
+  updateSelectedQueryIds(setQuery, query, selected);
```
markdboyd, line 36: If this really sucks or is a really bad way to do this, feel free to say so. It's just implementing a "nice to have" feature that isn't a core requirement of this ticket (selected granules on the overview page are automatically set as `ids` in the JSON request payload)

cypress/fixtures/seeds/collectionsFixture.json
```diff
@@ -13,7 +13,6 @@
       "version": "001",
       "createdAt": 1537830419369,
       "granuleId": "^.*$",
-      "provider_path": "granules/fake_granules",
```
markdboyd, line 16: This is for compatibility with the new API version which deprecated this field on collections

cypress/integration/bulk_granules_spec.js
```diff
@@ -0,0 +1,177 @@
...
+          });
+
+        cy.wait('@postBulkGranules');
+      });
+
+      it('show error message', () => {
+        cy.contains('.error__report', errorMessage);
+      });
+
+      it('hides error message when modal is closed and re-opened', () => {
```
markdboyd, line 113: I debated whether this test was a bit of overkill

app/src/css/_buttons.scss
jjmccoy: Thank you for refactoring the delete button styles.

jjmccoy:
UI/UX: Great execution of the new bulk granule modal workflow. LGTG once the tests pass.

app/src/js/components/Granules/bulk-granule-delete.js
dopeters: You should be able to just do something like this:

```
  const [query, setQuery] = useState(defaultQuery);

  const currentQuery = JSON.stringify({
    ...query,
    ids: selected
  }, null, 2);
```
And then pass `currentQuery` as the value for the `TextArea`
Because when `query` or `selected` changes, it'll cause the component to rerender and `currentQuery` will be updated

laurenfrederick:
I want @jjmccoy to make the final call. We had discussed and while they are similar it seemed confusing to have them both be one flow, since delete does not have a workflow and has the force remove from cmr setting.

markdboyd:
I had debated putting them into one component but with the instruction differences and different default queries, it felt easier to just keep them separate. But maybe it wouldn't be too bad to combine them. Happy to do it if that's the team preference

dopeters:
> I want @jjmccoy to make the final call. We had discussed and while they are similar it seemed confusing to have them both be one flow, since delete does not have a workflow and has the force remove from cmr setting.

To clarify, I think it makes sense to separate out the different flows, but the actual component that's used could potentially be a single one that has different props passed for each. I'm a fan of making things reusable where possible, but will defer to @jjmccoy 😄

jjmccoy:
I think we can leave the bulk delete and operations modal components separate for now and look at refactoring them later into one component if it is necessary. Sometimes it can get confusing and possibly run into code bloating if we consolidate components into one big component but I think if we utilize comments in the code more this could elevate comprehension and interpretation issues. For this ticket, I think we are good to go but if @markdboyd feels that it isn't much work for the points given and it's feasible to consolidate then I'm all for it.

PR #780, opened 2020-06-01
CUMULUS-1917 Add download report button

dopeters:
**Summary:** Adds a download report button to the Download Report column on the reconciliation report list page

Addresses [CUMULUS-1917: Reconciliation Reports - List - Update table to have Download button in table row](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1917)

## Changes

* Added the download button. Instead of using the table config, added it to the list page itself so that we can dispatch the `getReconciliationReport` action.
* `accessor` functions should return primitives, so where possible, when it was returning JSX, I used the `Cell` option instead
https://github.com/tannerlinsley/react-table/blob/master/docs/api/useTable.md#column-options

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### NOTE
~~Includes changes from https://github.com/nasa/cumulus-dashboard/pull/782, so should be merged after that~~


cypress/integration/reconciliation_spec.js
```diff
@@ -67,12 +67,14 @@ describe('Dashboard Reconciliation Reports Page', () => {
...
+      cy.get('.button__row--download').should('have.length', 3);
+    });
+
     it('deletes a report when the Delete button is clicked', () => {
       cy.visit('/reconciliation-reports');
-      cy.get('[data-value="inventoryReport-20200114T202529026"] > .td')
-        .eq(5)
-        .find('button')
-        .click();
+      cy.get('[data-value="inventoryReport-20200114T202529026"] > .td .button__row--delete').click();
```
jennyhliu, line 77: :+1:

app/src/js/utils/table-config/collections.js
```diff
@@ -12,8 +12,11 @@ import BatchDeleteWithGranulesContent from '../../components/DeleteCollection/Ba
 export const tableColumns = [
   {
     Header: 'Name',
-    accessor: row => <Link to={`/collections/collection/${row.name}/${row.version}`}>{row.name}</Link>,
-    id: 'name',
+    accessor: 'name',
```
jennyhliu, line 15: It seems making `accessor` primitive can solve https://bugs.earthdata.nasa.gov/browse/CUMULUS-1842.  I'm testing it out and may make changes in additional places.  Thanks!

app/src/js/reducers/reconciliation-reports.js
```diff
@@ -78,4 +86,7 @@ export default createReducer(initialState, {
   [CLEAR_RECONCILIATIONS_FILTER]: (state, action) => {
     state.list.params[action.paramKey] = null;
   },
+  [RECONCILIATION_DELETE]: createSuccessReducer('deleted'),
```
jennyhliu, line 89: Will these reducers change anything?

app/src/js/utils/table-config/collections.js
dopeters: Oh awesome, that actually makes a lot of sense!

app/src/js/reducers/reconciliation-reports.js
dopeters: I don't think so. I had added them earlier and forgot to remove them. They match the other delete reducers but they probably aren't necessary. I can get rid of them

dopeters: Done https://github.com/nasa/cumulus-dashboard/pull/780/commits/b0656ad4e29d4e09e4dc3059fc536111c137d0f5

PR #788, opened 2020-06-11
CUMULUS-1872 Granules timers

dopeters:
**Summary:** Addresses interval timers on granules pages - primarily `listWorkflows` calls that are used for updating the options passed to bulk actions

Addresses [CUMULUS-1872: Fix Timers - Granules page](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1872)

## Changes

* Removes the workflow timers on the granules page. Also fixes the default state for workflow
* Some linting tweaks
* Refactors components wrapped in `withQueryWrapper` => turns them into function components that call `onQueryChange` in a `useEffect` block to fix console error.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


app/src/js/components/Collections/granules.js
```diff
@@ -42,7 +42,7 @@ const CollectionGranules = ({
   const displayName = strings.granules;
   const collectionId = getCollectionId(params);
   const view = getView();
-  const [workflow, setWorkflow] = useState();
+  const [workflow, setWorkflow] = useState(workflowOptions[0]);
```
dopeters, line 45: Need to set initial state. Execute was failing if you didn't select a new workflow from the dropdown. This allows the first value to actually be used.

app/src/js/components/Granules/granule.js
```diff
@@ -119,34 +114,21 @@ class GranuleOverview extends React.Component {
...
+    this.loadGranule();
   }
 
-  fastReload () {
-    // decrease timeout to better see updates
-    this.reload(true, updateInterval / 2);
+  loadGranule () {
+    const { dispatch, match } = this.props;
+    const { granuleId } = match.params;
+    dispatch(getGranule(granuleId));
```
dopeters, line 131: Instead of triggering a faster reload, we're just dispatching the action directly on bulk action success

app/src/js/components/Granules/overview.js
```diff
@@ -63,24 +58,18 @@ class GranulesOverview extends React.Component {
...
+    this.queryMeta();
   }
 
   queryMeta () {
     const { dispatch } = this.props;
     dispatch(listWorkflows());
-    dispatch(getCount({
-      type: 'granules',
-      field: 'status'
-    }));
```
dopeters, line 83: This getCount dispatch is taken care of in the index.js file. No need for it here

jjmccoy:
UI/UX: Looks good here. I don't see any issues with the timers on the Granules page. Needs logic review. I'm curious in what re-ingesting a granule looks like.

app/src/js/components/Granules/index.js
```diff
@@ -24,12 +24,15 @@ const Granules = ({
   const AllGranulesWithWrapper = withQueryWrapper(AllGranules, queryOptions, setQueryOptions);
 
   function query () {
+    dispatch(listGranules(queryOptions));
+  }
+
+  useEffect(() => {
```
flamingbear, line 31: I need education here.
is this just dispatching a getCount when the dispatch object changes? what changes the dispatch object?

app/src/js/components/Granules/granule.js
flamingbear: that makes sense.

app/src/js/components/Granules/index.js
flamingbear: I may have answered my own question, does this mean it only runs once? (like listWorkflows() https://github.com/nasa/cumulus-dashboard/pull/788/files#diff-30c61d87d412853c9930566dd50f2c45R73). 
And that (listworkflows) probably is what's needed in collections/granules.?

app/src/js/components/Granules/overview.js
flamingbear: This might be out of scope, but we did discuss making the overview cards values match what's in the displayed table.  I don't think this getCount dispatch works as written, but could it be updated to do that?  but there might already be a ticket for that.
Screen Shot 2020-06-15 at 2 31 26 PM

app/src/js/components/Pdr/index.js
```diff
@@ -26,12 +26,15 @@ const Pdrs = ({
...
+  }
+
+  useEffect(() => {
     dispatch(getCount({
       type: 'pdrs',
       field: 'status'
     }));
-    dispatch(listPdrs(queryOptions));
-  }
+  }, [dispatch]);
```
flamingbear, line 37: again, generally only run this once?

app/src/js/components/Granules/index.js
dopeters: Right on both counts. Added to collections/granules here https://github.com/nasa/cumulus-dashboard/pull/788/commits/bbf214de2aeb9d0ea7ee8f5d4b1b269c1ac961e9

app/src/js/components/Pdr/index.js
dopeters: Correct

dopeters:
> This might be out of scope, but we did discuss making the overview cards values match what's in the displayed table. I don't think this getCount dispatch works as written, but could it be updated to do that? but there might already be a ticket for that.

@flamingbear Oh good call on this. I think it is out of scope, but is something to think about. Not seeing an existing ticket for it, so I made one.
https://bugs.earthdata.nasa.gov/browse/CUMULUS-2029

Though I could see the argument that those cards are showing everything, regardless of what filters are applied, so might be worth discussing with the whole team

flamingbear:
I'm still seeing weirdness around those granules under collections => MOD09GQ => granules.  and just directly with granules => execute. 

Is it possible this is fakeAPI related.  If you just reload the app, choose granules, select the first granule, and execute.  You get a body with undefined workflow and an error.
I have two quick movies. and I can't attach them to github.

app/src/js/components/Granules/overview.js
jjmccoy: > > This might be out of scope, but we did discuss making the overview cards values match what's in the displayed table. I don't think this getCount dispatch works as written, but could it be updated to do that? but there might already be a ticket for that.
> 
> @flamingbear Oh good call on this. I think it is out of scope, but is something to think about. Not seeing an existing ticket for it, so I made one.
> https://bugs.earthdata.nasa.gov/browse/CUMULUS-2029
> 
> Though I could see the argument that those cards are showing everything, regardless of what filters are applied, so might be worth discussing with the whole team

The ticket for this is CUMULUS-1868.

flamingbear:
Looks good to me. 👍

dopeters:
> I'm still seeing weirdness around those granules under collections => MOD09GQ => granules. and just directly with granules => execute.
> 
> Is it possible this is fakeAPI related. If you just reload the app, choose granules, select the first granule, and execute. You get a body with undefined workflow and an error.
> I have two quick movies. and I can't attach them to github.
> So I hope you can download them from here: [imgur.com/a/IsrBtYI](https://imgur.com/a/IsrBtYI)
> Nope, can't download them, hopefully you can read them, or I can slack them to you.

Ahh my bad, I should have tested the whole flow before pushing. So, the issue was that, while it's now setting the initial state, if you arrive at the page before `listWorkflows` has been dispatched, then `workflowOptions` would be empty, so the initial state is still null. Made a change so that when `workflowOptions` changes, we set `workflow` to the first value

PR #793, opened 2020-06-18
[CUMULUS-1882] Fix TEA Lambda metrics

bt6453:

Addresses [CUMULUS-1882: TEA Lambda Success Metrics broken on Dashboard homepage](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1882)

## Changes

* Fix ES query to search for updated logs to determine TEA metrics
* Update links to Kibana to account for new queries

## PR Checklist

- [x] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [ ] Integration tests


package-lock.json
```diff
@@ -20306,28 +20306,28 @@
           "dependencies": {
             "abbrev": {
               "version": "1.1.1",
-              "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
```
jennyhliu, line 20309: I wonder why `package-lock.json` is updated when there is no `package.json` change.
"resolved" should point to URI.

bt6453: Interesting ,I didn't catch this. I fixed it to match develop and I'm not sure why `npm install` is updating this either. I'll be more careful going forward.

PR #803, opened 2020-06-30
CUMULUS-2040 Fix rec report pagination

dopeters:
**Summary:** Add logic so not all pages are shown

Addresses [CUMULUS-2040: Update simple pagination to not display all page numbers when there are a large number of pages](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2040)

## Changes

* Added logic to check if a page number falls with in the desired range that we want to display.
* If it doesn't, for the first instance, we add `...`, otherwise, we just return nothing

Easy way to test if you don't have a lot of conflicts:
In `SortableTable,js`, add `pageSize: 1` to the initialState that passed to `useTable` (line 95)
It would look like this:

```
initialState: {
  hiddenColumns: initialHiddenColumns,
  pageSize: 1
}
```

Then, go to `http://localhost:3000/reconciliation-reports/report/inventoryReport-20200114T202529026` and click the CMR tab, which has 15 conflicts. There will be one per page for testing the functionality.

## PR Checklist

- [X] Update CHANGELOG
- [ ] Unit tests
- [X] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - **CUMULUS-2024**
   - Fix bug where new providers and collections were not being pulled in as options on the Add Rule page
 
+- **CUMULUS-2040**
+  - Fix reconciliation report pagination so that it does not display all pages when there are a large number of conflicts
+
 ## [v1.9.0]
 
 ### BREAKING CHANGES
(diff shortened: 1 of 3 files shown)


jjmccoy:
Truncation is nice. I'm seeing this on my 15" pagination_lg. We may want to look at possibly what it would look like at 13" (I think that's the smallest screen size we could support at this time).

jjmccoy:
Here's what the emulator shows for smaller screens.
Screen Shot 2020-06-30 at 1 03 33 PM

jjmccoy:
Updates look great! The functionality is nice.

Screen Shot 2020-06-30 at 1 46 09 PM
Screen Shot 2020-06-30 at 1 45 32 PM

PR #806, opened 2020-07-02
CUMULUS-1831 Fix modal workflows

dopeters:
**Summary:** Update batch async modal workflow so that modal closes when finished

Addresses [CUMULUS-1831: re-ingest of a granule that has a non-existent execution, launches banner but has wrong modal workflow.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1831)

## Changes

* Updated flow so that modal closes after success/error if `getModalOptions` function is not passed
* Fix issue where the inclusion of certain fields was causing bulk rule enable/disable to fail

## PR Checklist

- [X] Update CHANGELOG
- [ ] Unit tests
- [X] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+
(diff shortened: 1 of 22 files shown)


jjmccoy:
LGTG Nice fix to the reingest flow. 

Screen Shot 2020-07-09 at 3 27 35 PM

PR #811, opened 2020-07-20
CUMULUS-2029 Update overview tiles to match what is displayed in table

dopeters:

Addresses [CUMULUS-2029: Overview cards do not match what is displayed in the table](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2029)

## Changes

* Updated how stats reducer shims in empty count values.
* Added a reducer for storing the count to be used in the sidebar separately. That way, the sidebar will always represent the unfiltered count
* Updated the Overview component so it dispatches the filtered count instead of having the items to display passed in.
* Updated `getCount` calls so they include query params

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - **CUMULUS-2019**
   - Support partial search
 
+- **CUMULUS-2029**
+  - Overview tiles are updated to represent what is shown in the table
+
 ### Fixed
 
 - **CUMULUS-1815**
--- a/app/src/css/modules/_table.scss
+++ b/app/src/css/modules/_table.scss
@@ -179,20 +179,34 @@
       border-bottom: 0;
     }
   }
-}
 
-.overview-num__wrapper ul li, .overview-num__wrapper-home ul li{
-  display: inline-block
-}
+  ul li {
+    display: inline-block;
 
-.overview-num__wrapper ul li:first-child .num-status{
-  background-color: $light-green;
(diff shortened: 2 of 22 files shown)


jjmccoy:
UI/UX: The overview stats are matching the badge and what's in the table and what status they correlate with. I have one question about the providers page. Does the number correlate with the providers in that table or collections/granules that are associated with those providers? If it's collections/granules is 0 but ingesting providers is 2, then I wonder what the pattern here should be?

dopeters:
> UI/UX: The overview stats are matching the badge and what's in the table and what status they correlate with. I have one question about the providers page. Does the number correlate with the providers in that table or collections/granules that are associated with those providers? If it's collections/granules is 0 but ingesting providers is 2, then I wonder what the pattern here should be?

Hmm this is a good question. The providers schema doesn't include a status or associated granules https://github.com/nasa/cumulus/blob/master/packages/api/models/schemas.js#L567, so I'm wondering if we need overview cards there at all

dopeters:
After consulting in Slack, I have removed the overview from the Provider's page https://github.com/nasa/cumulus-dashboard/pull/811/commits/52f28d7dddca9d460d1168294078a224d2a62cb6

jjmccoy:
> After consulting in Slack, I have removed the overview from the Provider's page [52f28d7](https://github.com/nasa/cumulus-dashboard/commit/52f28d7dddca9d460d1168294078a224d2a62cb6)

Thank you for investigating this. Sounds good.

PR #812, opened 2020-07-23
CUMULUS-1836 Dropdown custom values

dopeters:
**Summary:** Replaces react-autocomplete with react-bootstrap-typeahead

Addresses [CUMULUS-1836: Update Autocomplete component to allow custom values](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1836)

## Changes

* Updated dropdown component to use [react-bootstrap-typeahead](https://github.com/ericgio/react-bootstrap-typeahead)
* Updated options that are passed into Dropdown to be an array of objects so that they work with the new component.
* When `paramKey === 'limit'`, we're setting the `allowNew` prop on `Typeahead` to `true`. This is what allows us to add custom values. It also works when a user comes in with a custom param in the url, e.g. `?limit=35`.

## PR Checklist

- [X] Update CHANGELOG
- [ ] Unit tests
- [X] Adhoc testing
- [X] Integration tests


jennyhliu:
I suspect that was introduced by https://github.com/nasa/cumulus-dashboard/pull/810. The `develop` branch has the same behavior.

dopeters:
Nice find! Yeah this was definitely a result of that previous ticket. Made a change where the middleware filters out `startDateTime` and `endDateTime`. Also added an ava test to check that. Additionally, updated the cypress test that was checking the datepicker result so that it is actually checking the right thing 😅 

Finally, noticed that the COUNT_SIDEBAR reducer wasn't accurately reflecting the datepicker, so made a tweak there.

cypress/integration/granules_spec.js
```diff
@@ -225,7 +225,9 @@ describe('Dashboard Granules Page', () => {
         .first().contains('li', 'Completed').contains('li', 7)
         .next().contains('li', 'Failed').contains('li', 2)
         .next().contains('li', 'Running').contains('li', 2);
-      cy.setDatepickerDropdown('Recent');
+      // cy.setDatepickerDropdown('Recent');
```
jennyhliu, line 228: Why is this commented out?  If `recent` is selected, should the original numbers get displayed?
I think it's better to test both datepicker w and w/o granules returned.  Since 0 granules could mean issues with the query itself.

app/src/js/components/DropDown/dropdown.js
```diff
@@ -1,143 +1,146 @@
...
+  Input,
+} from 'react-bootstrap-typeahead';
+
+function renderInput({ inputRef, referenceElementRef, ...inputProps }) {
   return (
-    <div className='autocomplete__menu' style={style} children={items} />
+    <Hint
+      shouldSelect={(shouldSelect, e) => {
+        // Selects the hint when the user hits the 'enter' key.
+        return e.keyCode === 13 || shouldSelect;
```
jennyhliu, line 20: When I type a number and enter, it doesn't work, until I select the number I typed in.

jennyhliu:
There is also an issue with Granules page, when add a `search` text, the granule shows up, but  the Granule Overview cards all have 0s.
Screen Shot 2020-07-29 at 10 06 33 AM
Screen Shot 2020-07-29 at 9 43 11 AM

cypress/integration/granules_spec.js
dopeters: Yeah that's a good point. Added to that test https://github.com/nasa/cumulus-dashboard/pull/812/commits/d551a5a0263e509e2d70554008281f3952c335db

app/src/js/components/DropDown/dropdown.js
dopeters: This was a little more involved because when adding a new option, there is no `<Hint>` component. That's why it wasn't working on enter in that case. 

As a workaround, I added an `onKeyDown` handler that fires when `allowNew === true` and `e.keyCode === 13`. It takes `e.target.value` and makes that the selection. Also had to add a ref to the <Typeahead> component so that the menu could be closed after this was called.

https://github.com/nasa/cumulus-dashboard/pull/812/commits/6d6cc851719f7ce83475edb7bdf1e47be511a4ae

dopeters:
Thanks for your eagle eyes here @jennyhliu! I think I covered everything and I've been testing as many cases as possible.

Re: the datepicker changing the sidebar numbers on the PDRs page: made a quick tweak that I had made on the granules page but hadn't made there https://github.com/nasa/cumulus-dashboard/pull/812/commits/9eb3b0694f172cbcca2a7c3db5b3d06d160600a9

Re: Overview card numbers when a search param is present: made a change so that the `filterQueryParams` function handles the `search` param and converts it to `infix`. https://github.com/nasa/cumulus-dashboard/pull/812/commits/0caadc46b9fffcd546257d4dec531313b4a6d576

I did notice one other small thing: If you have a small total number of records (say 5) and set the Results Per Page to a small number (say 2) and then change to page 3. You should see one record. However, if you then clear Results Per Page, it will show no results because it will still be querying page 3, but the limit is now 50 (so results will only be on page 1). I think it makes sense to tackle it in [CUMULUS-1892](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1892) since that deals with pagination, but let me know if you think it should be addressed here.

PR #817, opened 2020-07-29
release v1.10.0 redux : fix elliptical and dot-prop security fixes.

flamingbear:
elliptical could be fixed in place with the suggested .
dot-prop required ava upgrade to 3.11
as a result two tests had to be updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "@cypress/webpack-preprocessor": "^4.1.2",
     "audit-ci": "1.3.0",
     "autoprefixer": "^9.7.3",
-    "ava": "^2.4.0",
+    "ava": "^3.11.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
       "./test/_setup-browser-env.js",
       "ignore-styles"
     ],
+    "timeout": "3m",
```
flamingbear, line 71: This is a guess.  Something changed in AVAs timeout it was working fine on my local machine and not on the CircleCI build.

flamingbear: Just checked, in v2. there is no default timeout
https://github.com/avajs/ava/blob/v2.4.0/docs/07-test-timeouts.md#test-timeouts

PR #818, opened 2020-07-30
Fix execution search and add cypress test

dopeters:
**Summary:** Fixes executions search

## Changes

* Removed `paramKay` prop from `Search` component, which was causing the search to not work as expected
* Added cypress test for executions search

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


dopeters:
Seems like this is not actually broken on develop

PR #833, opened 2020-08-07
Naga1783

Nnaga1:
2046 - added api version on footer to be seen on all pages
2067 - changed CSS formatting of footer


app/src/css/_base.scss
```diff
@@ -42,6 +42,15 @@ body {
...
 }
 
+row {
+    flexDirection: 'row';
+    height: 50;
+    justifyContent: 'center';
+    alignItems: 'center';
+    position: 'absolute'; //Here is the trick
+    bottom: 0;
+}
```
dopeters, line 52: I don't think this is doing anything, since `row` is not a valid selector and some of the properties and invalid

app/src/js/components/Footer/footer.js
```diff
@@ -18,7 +20,10 @@ class Footer extends React.Component {
       <div className='footer'>
         <div className='api__summary'>
           { authenticated &&
-            <h5 className='api__version'>Cumulus API Version: { versionNumber }</h5>
+            <div>
+              <h5 className='dashboard_version'> Cumulus Dashboard Version: { DASH_VERSION } </h5>
```
dopeters, line 24: can we update the class with two underscores so it's consistent with `api_version`?
Also, can we update this so that it uses the same styles as `api_version`? 
```suggestion
              <h5 className='dashboard__version'> Cumulus Dashboard Version: { DASH_VERSION } </h5>
```

Basically would need to do this in _base.scss:

```
.api__version,
.dashboard__version{
  font-weight: 400;
  color: $white;
  padding: 10px 0 2px 0;
}
```

app/src/js/components/Footer/footer.js
```diff
@@ -18,7 +20,10 @@ class Footer extends React.Component {
...
+              <h5 className='api__version'> Cumulus API Version: { versionNumber } </h5>
```
dopeters, line 25: Looks like a test is failing due to the spacing around this
```suggestion
              <h5 className='api__version'>Cumulus API Version: { versionNumber }</h5>
```

Also, while we're at it, can we add a similar test for Dashboard version?

Nnaga1:
2046: Cumulus Dashboard Version showing on all screens
2067: Changed formatting of Footer

PR #834, opened 2020-08-10
CUMULUS-2090 Add ReportHeading component

dopeters:
**Summary:** Add a reusable ReportHeading component for use across all report types

Addresses [CUMULUS-2090: Reports: GNF View - Heading (Status & Download Feature)](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2090)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


dopeters:
> Works great! I have one comment: In the mockup for the ticket, the report header shows report type.
> Screen Shot 2020-08-11 at 2 29 23 PM

Oh good catch! I've updated our fixtures to include the "type" property and added "type" to the heading.

Quick note: Initially, I thought including the type also meant the "Granules Not Found 15" section at the bottom of the screenshot, but looking through the other mockups, I believe that is specific for the GNF report type. This I why I only added it to the main heading, but let me know if you think it should also be added below

jennyhliu:
Oh.  Sorry.  I didn't notice the report itself doesn't have `type` property, so we can't add it to fixture until the api/lambda gets updated.  So you could remove the updates for my comment for now.

dopeters:
> Oh. Sorry. I didn't notice the report itself doesn't have `type` property, so we can't add it to fixture until the api/lambda gets updated. So you could remove the updates for my comment for now.

Ahh ok wasn't sure if there was a good reason for the type not to be there, but that makes sense. Reverted the changes

PR #838, opened 2020-08-17
CUMULUS-2068 Dropdown update

dopeters:

Addresses [CUMULUS-2068: Dashboard: UI Interaction Fix - Dropdowns](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2068)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


jjmccoy:
Nice work. The interaction and dropdown look great! Thank you. I noticed some hover and selection wonkiness when it comes to styling. 

After selection and mouse moved away it still shows the hover state:
Screen Shot 2020-08-18 at 7 07 45 PM

For dropdown item hover: the blue browser default is overriding our custom link blue:
Screen Shot 2020-08-18 at 7 00 31 PM

dopeters:
@jjmccoy I added the prop `blurInputOnSelect={true}` to the Select component to unfocus the input after selecting the option.

I also updated the background color of hovering on the option. I also just made some updates so that some of the styles in SimpleDropdown are being pulled as variables from our css files so that future changes will be easier. Let me know what you think!

jjmccoy:
> @jjmccoy I added the prop `blurInputOnSelect={true}` to the Select component to unfocus the input after selecting the option.
> 
> I also updated the background color of hovering on the option. I also just made some updates so that some of the styles in SimpleDropdown are being pulled as variables from our css files so that future changes will be easier. Let me know what you think!

@dopeters Oh cool! Yeah trying to use CSS variables more. They make it so much better. Thanks.

I went back looked at the mockups and noticed the styling. My bad for not including this in the ticket as a reference for you.  Would it be too much trouble to make the inline hover #EBF3F8 for the background and we will keep the selected background color our `$ocean-blue`? Let me know if it's a bother. Also, I noticed that the dropdown option arrows are supposed to be white in all states -- it's a little hard to see the grey on blue contrast. 

Cumulus Dashboard - Home - Filters Shown

PR #839, opened 2020-08-24
Mhs/cumulus 1/fix random items

flamingbear:
**Summary:** three small changes

Addresses [CUMULUS-1961: Found working other tickets](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1961)

## Changes

* Makes ranges in reconciliation reports include time and date in UTC
* Prevents later render error when action.data.results is undefined
* Rewords the API version warning to be more clear about which versions need to be updated.

### DIFF
--- a/app/src/js/components/ReconciliationReports/report-heading.js
+++ b/app/src/js/components/ReconciliationReports/report-heading.js
@@ -35,11 +35,11 @@ const ReportHeading = ({
   ];
 
   const formattedStartTime = startTime
-    ? moment(startTime).utc().toLocaleString()
+    ? moment(startTime).utc().format('YYYY-MM-DD H:mm:ss UTC')
     : 'missing';
 
   const formattedEndTime = endTime
-    ? moment(endTime).utc().toLocaleString()
+    ? moment(endTime).utc().format('YYYY-MM-DD H:mm:ss UTC')
     : 'missing';
   return (
     <>
--- a/app/src/js/components/ReconciliationReports/report-heading.js
+++ b/app/src/js/components/ReconciliationReports/report-heading.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Dropdown as DropdownBootstrap } from 'react-bootstrap';
+import moment from 'moment';
 import Breadcrumbs from '../Breadcrumbs/Breadcrumbs';
 import ErrorReport from '../Errors/report';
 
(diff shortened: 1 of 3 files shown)


app/src/js/components/ReconciliationReports/report-heading.js
```diff
@@ -34,11 +35,11 @@ const ReportHeading = ({
   ];
 
   const formattedStartTime = startTime
-    ? new Date(startTime).toLocaleDateString()
+    ? moment(startTime).utc().toLocaleString()
     : 'missing';
 
   const formattedEndTime = endTime
-    ? new Date(endTime).toLocaleDateString()
+    ? moment(endTime).utc().toLocaleString()
```
dopeters, line 42: Does `toLocaleString()` still do anything with `moment.js`?

Also, this date formatting doesn't match the mockups, but not sure how important that is.

Screen Shot 2020-08-24 at 1 07 54 PM

[Mockup](https://xd.adobe.com/view/8a2afd0f-2e25-4814-79a3-11f3cc85c0c3-fc1b/screen/433a3107-824c-4120-a6bb-ce790791ca9d/)

flamingbear: Yeah, we can ask @jjmccoy, But since we have timestamps going into the reports, we need to have timestamps coming out.  
The toLocaleString() does appear to format 
```
 moment().utc().toLocaleString()
'Mon Aug 24 2020 17:20:33 GMT+0000'
```
But I didn't look at the mockup.  I can see now wanting the Day of Week and maybe not the TZ offset?

dopeters: Ah yeah, you're right it does. Though it seems like using `moment(startTime).utc()` without still works/displays a string. But probably best to have it.

Agreed re: the formatting and needing the timestamp. Definitely think we can nix day of the week though. And the TZ offset, as long as we're using UTC everywhere else (which I think we are?)

flamingbear: `(which I think we are?)` we should be....

I think moment(datething).utc() will create a utc moment object that is coerced into a string when we print it and we should probably be explicit.  I will find a better format string and let you and Juanisa decide if it's good enough.

jjmccoy: Hmm...

jjmccoy: Can we separate that please: The 'Date Range' and then under the 'Timestamp' and they will both be stacked in the same row with the report title? Please keep in mind they can select 'created', 'ingested', 'reingested', 'granule start time', or 'granule end time' as their timestamp as well.

flamingbear: I'm updating this with the format that got the most votes, we can address new inputs in an architecture meeting.

PR #842, opened 2020-09-09
CUMULUS-2087: Reconcilation Report Type selection

jennyhliu:

Addresses [CUMULUS-2087: Reports: Report Type Selection Page](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2087)

## Changes

* Add Reconciliation Report creation page with report type selection

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


dopeters (approved):
Looks great, works great! Some commented out that should be removed but shouldn't hold back approval.

My only concern was that currently Internal reports don't render anything when you click to them from the list page.
Screen Shot 2020-09-10 at 3 03 31 PM

However, after reading through [CUMULUS-2114](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2114) it says it should either download or be presented as JSON, so it probably makes sense to tackle that in that ticket.

PR #844, opened 2020-09-18
CUMULUS-1966 Create Rec Report Form

dopeters:

Addresses [CUMULUS-1966: Reports: Create Report - Inventory](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1966)

## Changes

* Package changes - added `final-form` and `react-final-form` to easily handle form state. Also updated cypress so that we can use "experimentalNetworkStubbing": true. This allows us to grab the request body from the create report form to ensure that it is formatted as expected
* Made Datepicker its own component so that single datepickers are reusable. Renamed the range to DatepickerRange. Also made a change to the DatepickerHeader to handle whether it should be visible or not.
* Create report form includes all fields that can be submitted to create a report. For collection id/granule id/provider, made it so when a value it added to one, the other two are disabled. These are set up as multi select dropdowns whih are passed to the API as an array
* Various style tweaks

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


jjmccoy (approved):
Great work on this especially on the "Additional Filters" section. LGTG!

PR #845, opened 2020-09-23
CUMULUS-2108: text area for meta on execute modal

nemreid:

Addresses [CUMULUS-2108: As an operator, I want to be able to submit bulk access constraint value updates via the dashboard.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2108)

## Changes

* Updated `Execute` modal for ApplyWorkflow with a collapsible textArea that contains a meta field for the API request updates in CUMULUS-2107 on granules overview page, granules list page, and granules details page.
* Added a unit test for the execute dialog return value.
* Updated bulk granules modal default text to contain a `meta` field to add to API request.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


app/src/js/utils/table-config/granules.js
```diff
@@ -112,6 +114,33 @@ export const errorTableColumns = [
...
+      options={config.options}
+      id={config.label}
+      onChange={config.selectHandler}
+    />
+    <Collapsible trigger='Add Custom Workflow Meta' triggerWhenOpen='Collapse'>
+      <form>
+        <TextArea
+          value={config.initialMeta}
+          id={`${config.label}-meta-text-area`}
+          onChange={config.metaHandler}
```
dopeters, line 134: I noticed when you completely clear out the meta text, the value changes to "workflow-meta-text area". Looks like this is only happening in places where it's directly using the useState setters, since there's no wrapping function to handle the id that's passed (so `Collections/granules` and `Granules/list`. Might be easier to just handle the id once right here to cover all cases with something like this:

```suggestion
          onChange={(id, value) => config.metaHandler(value)}
```

Up to you though on how you want to handle it!

dopeters (changes requested):
Looks really great! We do need to deal with the `onChange` weirdness that's happening on some of the pages. Also, can you please add a changelog entry? 😄

app/src/js/utils/table-config/granules.js
nemreid: That looks good. I pushed the requested updates.

dopeters (approved):
Thanks for the updates! Looks great!

PR #846, opened 2020-09-24
CUMULUS-2023:Update filter drop downs to have the same order as columns

jennyhliu:

Addresses [CUMULUS-2023: Search Module: Filter Ordering](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2023)

## Changes

* Update filter drop downs to have the same order as columns

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- **CUMULUS-2023**
+  - Update filter drop downs to have the same order as columns
+
 - **CUMULUS-2068**
   - Converts select dropdowns to react-select component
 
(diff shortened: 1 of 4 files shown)


jjmccoy:
@jennyhliu I have one question regarding 'Executions'. Is the 'Workflow' filter the same as 'Type' in the table? If so, we might want to rename the 'Type' column to 'Workflow' so there is better context for the user.

jennyhliu:
> @jennyhliu I have one question regarding 'Executions'. Is the 'Workflow' filter the same as 'Type' in the table? If so, we might want to rename the 'Type' column to 'Workflow' so there is better context for the user.

Yes.  Renamed.

jjmccoy:
> > @jennyhliu I have one question regarding 'Executions'. Is the 'Workflow' filter the same as 'Type' in the table? If so, we might want to rename the 'Type' column to 'Workflow' so there is better context for the user.
> 
> Yes. Renamed.

Awesome! Thanks for the clarification and update.

jjmccoy (approved):
LGTG! Thanks -- this will really help with continuity and context.

PR #847, opened 2020-09-24
fix granule modal kibana links to use anchor tags

markdboyd:


dopeters (approved):
😎 blinded by those front end skillz 🔥

PR #849, opened 2020-09-30
[CUMULUS-2021] Update search to use typeahead component

dopeters:

Addresses [CUMULUS-2021: Search Module: Search Input Fixes](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2021)

## Changes

* Updates the search component to use react-bootstrap-typeahead. This is the same package we're using for for the filter dropdowns, but here we're using the async version. This gives us autocomplete functionality as well as a clear button. 
* Also added a focus handler so that text gets highlighted when clicking into it.
* Adds additional params (and documentation for those params) to <Search> for easier use with the new component

To test: perform searches across the various app pages.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- **CUMULUS-2021**
+  - Update search to use react-bootstrap-typeahead to get autocomplete functionality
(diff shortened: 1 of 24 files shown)


jjmccoy (approved):
Awesome work on the feature request. This is going to make their lives so much easier. LGTG!

PR #861, opened 2020-10-16
Mhs/cumulus 2063/mmt links for private collections

flamingbear:
**Summary:** move all handling of MMT linking to the cumulus API

Addresses [CUMULUS-2063: private MMT Links](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2063)

## Changes

* removes all code related to fetching, creating and appending MMTLinks, and the MMTLink state.
* Updates calls to the /collection endpoint to include a request for MMTLinks
* Changes Table accessor from mmtLink to MMTLink to use the MMTLink data stored in the collection state. 
* Updates localAPI for testing to @cumulus/api@3.0.1-alpha.3

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [X] Integration tests


package.json
```diff
@@ -144,7 +144,6 @@
     "webpackbar": "^4.0.0"
   },
   "dependencies": {
-    "@cumulus/cmrjs": "1.11.0",
```
dopeters, line 147: 🙌

app/src/js/components/Collections/list.js
```diff
@@ -107,7 +102,7 @@ class CollectionList extends React.Component {
 
           <List
             list={list}
-            data={data}
+            data={list.data}
```
dopeters, line 105: Not a huge thing, but I believe we use list.data as the default for the data prop, so we should be able to exclude this prop

dopeters (approved):
Looks good to me! So much simpler not having to make the extra request client side.

PR #862, opened 2020-10-16
Remove the eslint plugin for webpack.

flamingbear:
We test the code during CI with an explicit eslint step. This just wastes time.

I used 'speed-measure-webpack-plugin' to determine that the startup time for
`npm run serve` went from ~40 sec to ~20 sec on my machine.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -50,12 +50,6 @@ const CommonConfig = {
                 '@babel/preset-react'
               ]
             }
-          },
-          {
-            loader: 'eslint-loader',
-            options: {
-              emitWarning: true
-            }
           }
         ]
       },


dopeters (approved):
👍

cypress/integration/collections_spec.js
```diff
@@ -172,17 +171,14 @@ describe('Dashboard Collections Page', () => {
...
       cy.wait('@getCollections');
       cy.get('.table .tbody .tr').should('have.length', 5);
 
       cy.contains('.table .tbody .tr a', name)
-        .should('have.attr', 'href', `/collections/collection/${name}/${version}`)
-        .click();
+        .then(($res) => {
+          expect($res).to.have.attr('href', `/collections/collection/${name}/${version}`);
+          cy.wrap($res).click();
+        });
```
flamingbear, line 181: I'm hoping this means that the same item will be checked for the href and then will be clicked. :shrug:

localAPI/docker-compose-cypress.yml
```diff
@@ -14,3 +14,5 @@ services:
       - NODE_ENV=test
       - LOCALSTACK_HOST=localhost
       - LOCAL_ES_HOST=localhost
+    # cargo cult guesses from https://github.com/cypress-io/cypress/issues/350
+    shm_size: '512MB'
```
flamingbear, line 18: I definitely saw some errors that could be related to shared memory size.

PR #866, opened 2020-10-29
CUMULUS-2170 Granule inventory param

dopeters:
**Summary:** Quick fix to pass `granuleId` instead of `granuleIds` as a param when creating a Granule Inventory report. Ties into the changes from here: https://github.com/nasa/cumulus-dashboard/pull/863

Addresses [CUMULUS-2170: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-XXX)

## Changes

* Pass `granuleId` instead of `granuleIds`

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/js/components/Granules/overview.js
+++ b/app/src/js/components/Granules/overview.js
@@ -150,7 +150,7 @@ class GranulesOverview extends React.Component {
     };
 
     if (granuleIds.length > 0) {
-      requestBody.granuleIds = granuleIds;
+      requestBody.granuleId = granuleIds;
     }
 
     this.setState({ isListRequestSubmitted: true });
--- a/app/src/js/components/Granules/overview.js
+++ b/app/src/js/components/Granules/overview.js
@@ -134,24 +134,13 @@ class GranulesOverview extends React.Component {
     const queryParams = this.generateQuery();
     const { collectionId, status, search: granuleIdFilter } = queryParams;
 
-    let granuleIds = selected;
-
-    // If there are no selected granules but a Granule ID is specified
-    // in the search filter, use only that.
-    if (granuleIds.length < 1 && granuleIdFilter) {
-      granuleIds = [granuleIdFilter];
-    }
-
     const requestBody = {
       reportName: listName,
       reportType: 'Granule Inventory',
       status,
-      collectionId
+      collectionId,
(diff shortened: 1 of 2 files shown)


app/src/js/components/Granules/overview.js
```diff
@@ -150,7 +150,7 @@ class GranulesOverview extends React.Component {
     };
 
     if (granuleIds.length > 0) {
-      requestBody.granuleIds = granuleIds;
+      requestBody.granuleId = granuleIds;
```
jennyhliu, line 153: line 142 doesn't seem to be correct: granuleIds = [granuleIdFilter];
I'm confused with using granuleIds here.

dopeters: Yeah I think the intention is to deal with passing a string vs. an array. However, since the api handles that anyway, do you think we should omit all of that and do something like this instead?

```
const requestBody = {
  reportName: listName,
  reportType: 'Granule Inventory',
  status,
  collectionId,
  granuleId: granuleIdFilter || selected
};
```

jennyhliu: Yup.  That's good.

dopeters: Cool, made that change

npauzenga: (chiming in because I worked on this) the only reason I didn't want to do `granuleId: granuleIdFilter || selected` was so that the API could always assume this property was an array, like Danielle said. If we don't do it that way it would need to accept either a string in cases where it's just one from the filter or an array in cases where we have multiple checked. 

A user could also just have one selected in the list, in which case you would have an array of one value.

jennyhliu: granuleIdFilter is not granuleId, so we have to distinguish string and array.  I suggest to add a comment to state that's actually a `search` infix.

dopeters: Alright, spoke with Nate and we agreed that this makes sense given the API can handle strings and arrays. Also added a comment as suggested to clarify what's being passed to that param.

jennyhliu (approved):
LGTG

PR #868, opened 2020-10-29
Mhs/cumulus 2242/deploy improvements prod build

flamingbear:
1. updates webpack.prod.js

   + public path is dependent on the input env SERVED_BY_CUMULUS_API when we
   are serving the dashboard from the api, we need to have a relative path to
   the bundles.  When we are behind cloudfront we need it to be root. This fixes **CUMULUS-2177**

   + explicitly turns off minimization. the minimizing was causing an error
   when you tried to delete a collection, there was no class to add a modal
   to.
   develop/app/src/js/components/DeleteCollection/DeleteCollection.js#L47

2. Changes selectors to work with mangled classes. The 'MenuList' class
   doesn't exist on the production bundle. So use a different selector.

3. Uses a production build in Docker and CI testing

4. Removes unnecessary validation-tests that really only tested how the
   application was deployed, not how it functioned.


jennyhliu (approved):
LGTM

PR #870, opened 2020-11-03
Test Fix 

flamingbear:
**Summary:** Don't depend on a previous month existing in the calendar, and just go back three months and grab the first date.


flamingbear:
closing since Jenny fixed in another PR

PR #871, opened 2020-11-04
Cumulus 2225 Add "Error Type" filter and sorting

npauzenga:

Addresses [CUMULUS-2225: Add filter for Error Type on Granule Errors Table](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2225)

Note that the tests are failing but it looks like the fix is in an open PR: https://github.com/nasa/cumulus-dashboard/pull/870

## Changes

* Renames "Type" column on Failed Granules table on both home and /granules/faied to "Error Type"
* Moves the "Error Type" column to the leftmost column
* Allows sorting on "Error Type" column
* Adds "Error Type" filter to Failed Granules page

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


app/src/js/components/DropDown/dropdown.js
```diff
@@ -111,7 +111,7 @@ const Dropdown = ({
   }, [dispatch, getOptions]);
 
   return (
-    <div className={`filter__item form-group__element filter-${paramKey}`}>
+    <div className={`filter__item form-group__element filter-${paramKey.includes('.') ? paramKey.split('.')[0] : paramKey}`}>
```
npauzenga, line 114: This is because we might have child properties as the paramKey. In this case it would be `error.Error`, which we wouldn't want as a class name.

app/src/js/components/Granules/list.js
```diff
@@ -74,6 +77,16 @@ const AllGranules = ({
...
 
+  useEffect(() => {
+    dispatch(
+      getCount({
+        type: 'granules',
+        field: 'error.Error.keyword',
+        sidebarCount: false
+      })
+    );
+  }, [dispatch]);
```
npauzenga, line 88: Originally I was thinking we'd need a new API to get the list of granule errors. I think we can just use the existing /stats/aggregates endpoint though. I'm bucketing the aggregates by the error type, then just getting a list of all of those buckets (ignoring the actual count).

cypress/fixtures/seeds/granulesFixture.json
```diff
@@ -209,8 +209,8 @@
       "execution": "https://console.aws.amazon.com/states/home?region=us-east-1#/executions/details/arn:aws:states:us-east-1:123456789012:execution:TestSourceIntegrationIngestGranuleStateMachine-Rcei43QwgSP6:ffdb0984-c64e-4b1d-8af0-81938ec8acf9",
       "published": false,
       "error": {
-        "Cause": "{\"errorMessage\":\"Source file not found s3://cumulus-test-sandbox-internal/non-existent-path/non-existent-file\",\"errorType\":\"FileNotFound\",\"stackTrace\":[\"S3Granule.sync (/var/task/index.js:147549:13)\",\"<anonymous>\",\"process._tickDomainCallback (internal/process/next_tick.js:228:7)\"]}",
-        "Error": "FileNotFound"
+        "Cause": "{\"errorMessage\":\"verifyFile s3://cumulus-test-sandbox-internal/non-existent-path/non-existent-file failed: Actual file size 2MB did not match expected file size 3MB\",\"errorType\":\"UnexpectedFileSize\",\"stackTrace\":[\"S3Granule.sync (/var/task/index.js:147549:13)\",\"<anonymous>\",\"process._tickDomainCallback (internal/process/next_tick.js:228:7)\"]}",
+        "Error": "UnexpectedFileSize"
```
npauzenga, line 213: Flipping this just to be able to test different error types.

jjmccoy:
Nice work! LGTG once tests pass on CircleCI.

jjmccoy (approved):

PR #873, opened 2020-11-05
Bugfix: Update error name in tests. 

npauzenga:

In a previous PR I changed the type of errors returned in our fixture. They were previously the same. The trick in this test is that the order of the returned granules might change and the test might fail. We don't actually care about the order here, just that the two granules are listed.

## Changes

* Updates 1 spec

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


dopeters (approved):
👍 LGTM

PR #879, opened 2020-11-11
[CUMULUS-2238] Fix datepicker CSS across browsers

npauzenga:
**Summary:** 

Addresses [CUMULUS-2238: Dashboard: Date Time Picker CSS Fix](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2238)

## Changes

* Adds CSS rules to the DatePicker CSS for our React-Datetime-Picker container
* Overrides CSS rules for the React-Datetime-Picker component

Previously we had the following problems:
* Chrome did not show the carrot on the AM/PM dropdown
* Firefox showed the default system styling for the AM/PM dropdown
* Clicking on or updating the date or time would cause all of the fields (Duration, Start Date, Time Format, etc.) to shift slightly

To stop the elements shifting, I switched the font to Monospace. The problem here was that the digits in the date were taking less horizontal space than the placeholder characters. This is definitely a stylistic change so worth reviewing from a design perspective. There is also now a hard `width` on the inputGroup container to ensure it doesn't move.

The carrot issue in Firefox is a longstanding Firefox issue. The fix here is a hack that strips all style via the `appearance` property and then, if the browser is firefox, places a new carrot icon. Not ideal and I'm definitely open to suggestion but I'm not sure there's another way to do this without modifying the markup (which would be tough because it's in the React DateTime Picker library).

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


jjmccoy:
@npauzenga Reviewed dtp module in Chrome and Firefox. The Monospace looks fine with our other font (which is sans-serif). I noticed that the dtp module in the header has extra spacing between the inputs and dropdowns.

**Chrome**
Screen Shot 2020-11-12 at 4 04 18 PM
Screen Shot 2020-11-12 at 4 04 45 PM

**Firefox**
Screen Shot 2020-11-12 at 4 06 23 PM
Screen Shot 2020-11-12 at 4 06 05 PM

npauzenga:
Thanks @jjmccoy! Great catch, that is definitely not right.

I just pushed a change that wraps the DatePicker object in a new div with a different fixed width for each instance. I hate to add markup for something like this but I have a feeling this might be the cleanest solution. Our previous implementation didn't have this issue because we used fixed margins to get the whole DatePicker component to the size we wanted. Removing that means we have to constrain it some other way. I blame Firefox.

Totally open to other solutions if you have a preference.

jjmccoy:
> Thanks @jjmccoy! Great catch, that is definitely not right.
> 
> I just pushed a change that wraps the DatePicker object in a new div with a different fixed width for each instance. I hate to add markup for something like this but I have a feeling this might be the cleanest solution. Our previous implementation didn't have this issue because we used fixed margins to get the whole DatePicker component to the size we wanted. Removing that means we have to constrain it some other way. I blame Firefox.
> 
> Totally open to other solutions if you have a preference.

@npauzenga I blame Firefox too! Haha. This looks good. Yeah, I think this is the cleanest markup solution.

jjmccoy (approved):
LGTG!

PR #880, opened 2020-11-11
CUMULUS 2206 Performance updates

dopeters:
**Summary:** Improve component performance

Addresses [https://bugs.earthdata.nasa.gov/browse/CUMULUS-2206: Investigate event handler warnings in Chrome](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2206)

## Changes

* Biggest change is lazy loading some components, most notably the `<SortableTable>` within the main `<List>`
  - This also required adding `babel-eslint` to correct handle the syntax
* Couple of small CSS tweaks that avoid absolute positioning, which can cause layout shifts (which have a slight impact on performance/UX)
* Removed FontAwesome imports from App.js as this is not necessary for the CSS rendered icons as we are already including the js file in the html template and using both causes slower loading. Changed instances using <FontAwesomeIcon> to import the specific icon.
* Changed <List> to be a function component. This also revealed some extraneous props like sortId and order that I was able to get rid of since they are managed by react-table.

Overall, we can see specific performance increase in the loading of components with far fewer components taking longer than 1.5ms to load. (this is from the granules page which tends to take the longest). Though there is still the occasional warning in Chrome, I've seen far fewer, and this will allow us to leverage lazy loading elsewhere in the app going forward.

Before Lazy loading:
Screen Shot 2020-11-12 at 4 59 59 PM

After lazy loading
Screen Shot 2020-11-12 at 5 01 05 PM

## PR Checklist
- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


markdboyd:
I'm just here to 👏 for this. Well done. Also, React has a component named `Suspense`? How dramatic

jjmccoy:
UI/UX Review: Great work! Refactoring of FontAwesome and UI/CSS components were needed.

jjmccoy:
Team: This still needs a technical review especially regarding the performance.

app/src/js/components/SortableTable/SortableTable.js
```diff
@@ -35,18 +35,16 @@ IndeterminateCheckbox.propTypes = {
...
+  clearSelected,
   data = [],
+  initialHiddenColumns = [],
   onSelect,
-  clearSelected,
+  rowId,
+  shouldManualSort = false,
   shouldUsePagination = false,
-  initialHiddenColumns = []
+  tableColumns = [],
```
flamingbear, line 47: This looks like you removed sortId and order, in favor of always having a sortBy but I see sortId still used on <SortableTable’s.  execution-events.js   Is that an oversight, or does it get passed down somehow to another component?

app/src/js/components/Table/Table.js
```diff
@@ -8,224 +8,209 @@ import omitBy from 'lodash/omitBy';
...
-
+const List = ({
+  action,
+  bulkActions,
+  children,
+  data,
+  dispatch,
+  filterAction,
+  filterClear,
+  initialSortId,
```
flamingbear, line 29: I'm not seeing sortId as an prop for List anymore, but I do see it though out the code. That seems wrong.

app/src/js/components/Table/Table.js
```diff
@@ -8,224 +8,209 @@ import omitBy from 'lodash/omitBy';
...
+  useEffect(() => {
     // Remove parameters with null or undefined values
-    const params = omitBy(list.params, isNil);
+    const newParams = omitBy(list.params, isNil);
 
-    if (!isEqual(params, this.state.params)) {
-      // eslint-disable-next-line react/no-did-update-set-state
-      this.setState({ params }, () => this.setState({
-        queryConfig: this.getQueryConfig(),
+    if (!isEqual(newParams, params)) {
```
flamingbear, line 73: do we need this test? wont the useEffect dependencies handle that?   Or is this special because some params can become Null/undefined?

flamingbear (changes requested):
I think I basically have one question regarding the refactor but it appears a couple of times.

app/src/js/components/SortableTable/SortableTable.js
dopeters: Yeah this was an oversight. The idea is that when using a `<SortableTable>` via `<List>`, we're doing all of the request for the data and sorting server side (aka manual sorting). The `initialSortId` prop on `<List>` allows us to set up the sort_key for the API request, but then all of the rest of the sorting will be handled based on the user's interaction with the table, so the props are never updated/used anyway.

On the other hand, when using a plain `<SortableTable>`, we're leveraging the react-table's automatic/internal sorting feature, so we also do not need to pass in props for that.

app/src/js/components/Table/Table.js
dopeters: Another oversight, updated the props to be initialSortId instead when rendering <List>

app/src/js/components/Table/Table.js
dopeters: Yeah exactly, the useEffect will be triggered if values in list.params are null, but we only want to update the params if the non-null values change. Truthfully, not 100% sure if keeping track of the params like this is necessary, but was afraid to remove it, so wrote it in a way that would mirror the previous functionality.

flamingbear (approved):
Looks like the issues were cleaned up, and this is a nice improvement. A follow up ticket is created to implement the webpack changes necessary for lazy loading to provide significant value.

PR #881, opened 2020-11-12
CUMULUS-2135: Dashboard - Table Header Pagination Style Tweaks

jjmccoy:
**Summary:** 

Addresses [CUMULUS-2135: Dashboard: Table Header Pagination Style Tweaks](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2135)

## Changes

* Pagination header: Add` flex` width and basis in thirds and justified properly for even distribution of items [records, pagination-dropdown, options]. This helps to stop elements from shifting especially when # records display table records that are selected. 
* Pagination in table header and footer: Make sure that `text-align` is center. This ensures that both `pagination-dropdown ol` and `pagination-list ol` are centered in the `div`. 
* CSS Refactor: Removed duplicate pagination styling in `_table.scss` and added updates to the bootstrap overrides `_pagination.scss` for pagination styling. 

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


npauzenga:
Looks good @jjmccoy! I like the centered digits. Much cleaner.

Two quick things:

1. Should the top and bottom page navigation page navigation stay centered? It looks like the top row (Total Records, Previous/Next, Show x Per Page) scales appropriately but the bottom navigation stays put when you resize the window, causing the two to be out of alignment:

Screen Shot 2020-11-12 at 12 33 17 PM

2. This is probably out of scope for this ticket and may not even be something we can fix so, you know, take it with a grain of salt. In Chrome when I click on the page input field on the top page navigation, I'm presented with a dropdown of my recent entries. Same on the "Show X Per Page" input. I'm wondering if we can disable this? It seems mildly confusing because it looks like a dropdown but only shows the current page as an option. It looks like the element has `autocomplete=off`, which is what I would expect disables this so it's possible the browser will always show this based on the user's setting:

Screen Shot 2020-11-12 at 12 36 24 PM

jjmccoy:
> Looks good @jjmccoy! I like the centered digits. Much cleaner.
> 
> Two quick things:
> 
> 1. Should the top and bottom page navigation page navigation stay centered? It looks like the top row (Total Records, Previous/Next, Show x Per Page) scales appropriately but the bottom navigation stays put when you resize the window, causing the two to be out of alignment:
> 
> Screen Shot 2020-11-12 at 12 33 17 PM
> 
> 1. This is probably out of scope for this ticket and may not even be something we can fix so, you know, take it with a grain of salt. In Chrome when I click on the page input field on the top page navigation, I'm presented with a dropdown of my recent entries. Same on the "Show X Per Page" input. I'm wondering if we can disable this? It seems mildly confusing because it looks like a dropdown but only shows the current page as an option. It looks like the element has `autocomplete=off`, which is what I would expect disables this so it's possible the browser will always show this based on the user's setting:
> 
> Screen Shot 2020-11-12 at 12 36 24 PM

npauzenga:
Great, thanks @jjmccoy! Looks great to me! :shipit:

npauzenga (approved):
👍  LGTM!

PR #891, opened 2020-11-20
[CUMULUS-2262] Updated Granule Delete UX

npauzenga:
**Summary:** 

Addresses [CUMULUS-2262: Dashboard: Delete Granule WorkFlow](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2262)

## Changes

* Selecting published granules and clicking "Delete" will now notify the user that they will be removed from CMR and _then_ deleted in one step. Previously they would get an error saying it needed to be remove from CMR first.
* Selecting a mix of published and unpublished granules and pressing "Delete" and "Confirm" on the modal will remove and delete those that are published and just delete those that aren't.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


app/src/js/actions/index.js
```diff
@@ -383,6 +383,41 @@ export const deleteGranuleClearError = (granuleId) => ({
...
+          if (!errorPresent(removeResponse, types.GRANULE_DELETE_ERROR)) {
+            return dispatch(deleteGranule(granuleId));
+          }
+        }
+      );
+  }
+
+  // If this granule is NOT published, just delete it
+  return dispatch(deleteGranule(granuleId));
+};
```
npauzenga, line 419: Totally open to handling this series of synchronous calls differently. I'm replicating the approach I would take using async/await but there may be a better Redux/Thunk approach that I'm not seeing.

jjmccoy: @dopeters What do you think of this approach?

dopeters: I think this approach makes sense and follows the pattern we are using elsewhere. No issues from my perspective!

jjmccoy (approved):
LGTG!

PR #894, opened 2020-11-30
CUMULUS-2263 Clear pagination input on focus for cleaner UX

npauzenga:

Addresses [CUMULUS-2263:Dashboard: Table Header Input Interaction Tweaks](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2263)

## Changes

* Updates the Pagination Typeahead input to clear when focused
* Populates the Pagination dropdown list with the appropriate page numbers

Previously a user would click the pagination input and a dropdown would appear showing only the number entered (or any page numbers starting with that number). This is a little confusing because you would either expect to see no dropdown (text input only) or a dropdown containing all available page options.

I opted to keep the dropdown but try to make it useful. Now, when a user clicks the input, the field will clear and show all available pages in the dropdown. You can select one or type any number. This way, if there are 100s of pages, the typeahead might come in handy.

Totally open to feedback if there's a preference for hiding the dropdown in this case instead.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - **CUMULUS-2072**
   - Add 'Show More' and 'Show Less' buttons to error report
 
+- **CUMULUS-2263**
+  - Update Pagination input to show possible page options in dropdown
+
 ### Added
 - **CUMULUS-1895**
   - Update execution events to display more details in modal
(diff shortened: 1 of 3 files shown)


jjmccoy:
Looks good. Like that the dropdown allows the user to jump to the page especially if there are a lot of pages.

jjmccoy (approved):
LGTG!

PR #897, opened 2020-12-01
CUMULUS-2271 Allow horizontal scrolling on table columns

npauzenga:

Addresses [CUMULUS-2271: Dashboard: Table Column Horizontal Scroll
](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2271)

## Changes

* Shows the overflow on table cells when hovered

Previously the only way to show the full contents of a cell was to expand the cell's width by dragging the header. You can still do that but now, if there is overflow, you can hover on a cell and a scrollbar will appear.

Note that this removes the ellipses from truncated text. We can keep it and show the "..." when text is truncated, then remove the ellipses on hover BUT then the issue is what happens when the user moves their mouse outside of the cell? I'm not sure there's a great way to keep the full, non-truncated text in that case. We can of course handle it in JS but I'm wary of adding any handlers on a table with potentially hundreds of rows.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/css/modules/_table.scss
+++ b/app/src/css/modules/_table.scss
@@ -37,6 +37,22 @@
         &:hover {
           overflow: auto;
         }
+
+        /*
+         * From https://stackoverflow.com/questions/7492062/css-overflow-scroll-always-show-vertical-scroll-bar/10100209#10100209
+         * The horizontal scrollbar does not always show, depending on the user's OS settings.
+         * This forces a bar to render.
+        */
+        &::-webkit-scrollbar {
+          -webkit-appearance: none;
+          height: 8px;
+        }
+
+        &::-webkit-scrollbar-thumb {
+          border-radius: 4px;
+          background-color: rgba(0, 0, 0, .5);
+          box-shadow: 0 0 1px rgba(255, 255, 255, .5);
+        }
       }
       .table__main-asset {
         font-weight: $base-font-semibold;
--- a/app/src/css/modules/_table.scss
+++ b/app/src/css/modules/_table.scss
@@ -34,10 +34,10 @@
         padding: 1em 2em;
         vertical-align: middle;
         overflow: hidden;
-        text-overflow: ellipsis;
-        /*&:first-of-type {
-          padding: 1em 0 1em 2em;
-        }*/
+        min-height: 65px;
+        &:hover {
+          overflow: auto;
+        }
       }
       .table__main-asset {
         font-weight: $base-font-semibold;
--- a/app/src/css/modules/_table.scss
+++ b/app/src/css/modules/_table.scss
@@ -34,7 +34,6 @@
         padding: 1em 2em;
         vertical-align: middle;
         overflow: hidden;
-        min-height: 65px;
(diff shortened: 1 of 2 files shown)


app/src/css/modules/_table.scss
```diff
@@ -34,10 +34,10 @@
         padding: 1em 2em;
         vertical-align: middle;
         overflow: hidden;
-        text-overflow: ellipsis;
-        /*&:first-of-type {
-          padding: 1em 0 1em 2em;
-        }*/
+        min-height: 65px;
```
npauzenga, line 37: We might not need to add a min-height here if we don't want to. I have it because without it the row will expand vertically when hovered to fit the scrollbar. It does increase all of the row heights though. If we remove it the row will expand on hover, which also might not be terrible. I could go either way.

jjmccoy:
This is great @npauzenga! Is there a way to show the bar `on hover` for some reason it only shows when I start scrolling sideways on my trackpad?

app/src/css/modules/_table.scss
jjmccoy: @npauzenga I'm curious what the second option would look like with expand `on hover`.

npauzenga: @jjmccoy yeah, if you comment out this line (37) and hover over the table cells, they will expand to fit the scrollbar. Happy to jump on a screenshare if that's helpful.

npauzenga:
> This is great @npauzenga! Is there a way to show the bar `on hover` for some reason it only shows when I start scrolling sideways on my trackpad?

@jjmccoy Nice catch! So we've talked a bit about this and found that it's an OS setting. I checked with Danielle this morning and she saw the same thing you were. It appears to be an issue when you use a trackpad. In your case it sounded like plugging in a mouse didn't help, in Danielle's case it did, and in my case I could replicate the issue either way without changing OS settings.

The fix I just pushed forces a scrollbar to show. I was able to replicate the issue you saw by switching my OS "Show Scroll Bars" setting to "When scrolling" and this fix seems to do it. Would you mind checking again on your end?

jjmccoy:
> > This is great @npauzenga! Is there a way to show the bar `on hover` for some reason it only shows when I start scrolling sideways on my trackpad?
> 
> @jjmccoy Nice catch! So we've talked a bit about this and found that it's an OS setting. I checked with Danielle this morning and she saw the same thing you were. It appears to be an issue when you use a trackpad. In your case it sounded like plugging in a mouse didn't help, in Danielle's case it did, and in my case I couldn't replicate the issue either way without changing OS settings.
> 
> The fix I just pushed forces a scrollbar to show. I was able to replicate the issue you saw by switching my OS "Show Scroll Bars" setting to "When scrolling" and this fix seems to do it. Would you mind checking again on your end?

@npauzenga Thanks for investigating this. It shows `on hover` now. Yay!

jjmccoy (approved):
LGTG!

PR #898, opened 2020-12-01
[CUMULUS-1949] Modal Updates

dopeters:
**Summary:** Refactor modals. A lot of the changes were already there, so this was primarily adding the progress circle.

Addresses [CUMULUS-1949: Dashboard: Modal UI Refactor - Implementation](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1949)

## Changes

* Add `react-circular-progressbar` component to use for batch async modal progress indication
* Changes to ensure that modals will show success or errors for batch async commands
* Updates to async command modals to streamline modal presentation and show success where needed
* Modal titles will match button text with `confirm` function results displaying within the modal body so that user keeps context throughout modal journey

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


jjmccoy:
@dopeters Looks good! The progress animation is really nice! I noticed a few things. 

Progress Modal: I don't know with our app if we can add a delay before the action after the user clicks "confirm/send" for the progress message (One moment...) but if not it's fine. For those especially long lists for reingest, delete, etc. where the user just sitting there does the progress message stay up?

This `warning modal` is missing the `warning alert banner`:
Screen Shot 2020-12-03 at 11 16 30 AM

This modal has a weird gray background behind the modal body text. It looks like it is affecting the modals with singular selections except Collections for some reason 🤷‍♀️:

Screen Shot 2020-12-03 at 11 42 27 AM
Screen Shot 2020-12-03 at 11 43 02 AM

I think we can use Bootstrap's different size modals based on the content. This could be a `modal-lg`for `error modals` with details:
Screen Shot 2020-12-03 at 11 30 51 AM

This modal needs to be centered with the overlay and not the page:
Screen Shot 2020-12-03 at 11 31 51 AM

dopeters:
Thanks for all the feedback @jjmccoy!

- Progress modal: Wasn't finding a good way to add a delay (don't love adding unneeded timeouts, etc), but I did add the "One moment..." text that was visible in the mockup with the progress circle. Also tweaked the colors a bit to better match the mockup. The message should stay up for the entire time while the indicator updates.

- Warning modal: Added the alert here. Also updated all of the similar standalone modals to have alerts on success/error as needed. Updated the title as well

- Body text background: I had put that in an Alert component, but definitely unneeded. Removed.

- Modal size: I made size a prop on Default Modal so that can be dynamically changed now. Made the reingest modal `lg` as suggested.

Definitely realizing that's there probably a bigger effort here involving truly consolidating and standardizing all of our modals. We should think about how to break that into smaller pieces. Maybe building out specific modal types as their own tickets and then addressing each individual modal one by one until we've gotten everything?

Maybe we can even pass a prop to DefaultModal that indicates if it's success/error/etc and then it will display the proper alerts/content that way? Probably a lot of different options.

jjmccoy:
> Thanks for all the feedback @jjmccoy!
> 
> * Progress modal: Wasn't finding a good way to add a delay (don't love adding unneeded timeouts, etc), but I did add the "One moment..." text that was visible in the mockup with the progress circle. Also tweaked the colors a bit to better match the mockup. The message should stay up for the entire time while the indicator updates.
> * Warning modal: Added the alert here. Also updated all of the similar standalone modals to have alerts on success/error as needed. Updated the title as well
> * Body text background: I had put that in an Alert component, but definitely unneeded. Removed.
> * Modal size: I made size a prop on Default Modal so that can be dynamically changed now. Made the reingest modal `lg` as suggested.
> 
> Definitely realizing that's there probably a bigger effort here involving truly consolidating and standardizing all of our modals. We should think about how to break that into smaller pieces. Maybe building out specific modal types as their own tickets and then addressing each individual modal one by one until we've gotten everything?
> 
Yes, this sounds like a good plan. This effort was a quick refactor for layout and context. 

> Maybe we can even pass a prop to DefaultModal that indicates if it's success/error/etc and then it will display the proper alerts/content that way? Probably a lot of different options.

@dopeters Thanks for the updates. Looks great!

jjmccoy (approved):
LGTG! Nice work!

PR #900, opened 2020-12-03
[CUMULUS 2071] Update card interactions

npauzenga:

Addresses [CUMULUS-2071: Dashboard: UI Interaction Fix - Cards](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2071)

## Changes

* Replaces the static Granule Updates stats on the home page with clickable cards
* Animates the Reconciliation Report > Bucket Status cards

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/js/components/home.js
+++ b/app/src/js/components/home.js
@@ -97,7 +97,7 @@ class Home extends React.Component {
       <section className='page__section'>
         <div className='row'>
           <div className='heading__wrapper'>
-            {header}
+            <h2 className='heading--medium heading--shared-content--right'>{header}</h2>
           </div>
           <div className="overview-num__wrapper overview-num__wrapper-home">
             <ul id={listId}>
@@ -205,22 +205,9 @@ class Home extends React.Component {
             </div>
           </section>
 
-          {this.renderButtonListSection(
-            overview,
-            <h2 className='heading--medium heading--shared-content--right'>Updates</h2>
-          )}
-
-          {this.renderButtonListSection(
-            distErrorStats,
-            <h2 className='heading--medium heading--shared-content--right'>Distribution Errors</h2>,
-            'distributionErrors'
-          )}
-
-          {this.renderButtonListSection(
-            distSuccessStats,
-            <h2 className='heading--medium heading--shared-content--right'>Distribution Successes</h2>,
-            'distributionSuccesses'
-          )}
+          {this.renderButtonListSection(overview, 'Updates')}
+          {this.renderButtonListSection(distErrorStats, 'Distribution Errors', 'distributionErrors')}
+          {this.renderButtonListSection(distSuccessStats, 'Distribution Successes', 'distributionSuccesses')}
 
           <section className='page__section update--granules'>
             <div className='row'>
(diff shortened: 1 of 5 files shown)


app/src/js/components/home.js
```diff
@@ -23,7 +23,6 @@ import {
   seconds
 } from '../utils/format';
 import List from './Table/Table';
-import GranulesProgress from './Granules/progress';
```
npauzenga, line 26: This component is almost deprecated. It looks like it's still used on the PDRs page but I don't actually see it rendering. Could it be removed there (/components/Pdr/pdr.js)? If so we can delete the whole thing.

dopeters: Looks like that component is on the PDR page `/pdrs/pdr/MOD09GQ_1granule_v3.PDR`, so I think we should keep it for now. (Unless those need to be updated with card interactions, not sure)
Screen Shot 2020-12-03 at 2 05 44 PM

app/src/js/components/home.js
```diff
@@ -193,9 +218,10 @@ class Home extends React.Component {
               <div className="heading__wrapper">
                 <h2 className='heading--medium heading--shared-content--right'>{strings.granules_updated}<span className='num-title'>{numGranules}</span></h2>
               </div>
-
-              <GranulesProgress granules={granuleStatus} />
             </div>
+
+            {this.renderButtonListSection(updated, '')}
```
dopeters, line 223: I think we should either pass the header to this function or find a way to not render one if this is blank. Right now we're rendering two `heading__wrapper` divs

Also, unrelated, but I noticed Helmet was being rendered in that `renderButtonListSection` so it's on the page multiple times

If you're already making a change, I would suggest moving that into the main render function so we only have it once.

dopeters (changes requested):
Just a small suggestion re: the renderButtonListSection` function, but otherwise looks great! The addition of the rec report animation is awesome

app/src/js/components/home.js
npauzenga: Oh it's on the individual PDR page, got it. Hmm yeah, I would think we want these to be the individual cards as well. If you click on one it should bring you to the appropriate granules page but in this case also filtered somehow? I was thinking it would just be filtered by the collection but it looks like that's not quite right.

@jjmccoy what do you think? Danielle's route above has valid data: `/pdrs/pdr/MOD09GQ_1granule_v3.PDR`

Might need to be a separate ticket to swap these out and remove the `Progress` component.

jjmccoy: @npauzenga Oh okay. Hmm. I didn't realize those cards were associated with the PDR route. Yeah, that change might be out of scope for this ticket. This was supposed to be a quick UI interaction refactor.

app/src/js/components/home.js
npauzenga: Thanks @dopeters! I just pushed a change that updates `renderButtonListSelection` to accept a complete header element rather than text. It's not the DRYest solution but it might be the clearest. Lemme know if you'd prefer something different.

I also moved `Helmet` to the main render function. It looks like it doesn't totally matter where it is but I put it in the main `page__home` container at the top level.

app/src/js/components/home.js
npauzenga: Ok cool. It's not terribly complicated but it's a little different because we need to filter the Running/Completed/Failed results. If we can do that in a separate ticket that would be great.

dopeters (approved):
Looks awesome! Thanks for the tweaks.

PR #911, opened 2020-12-18
[CUMULUS-1885] Expand columns on double click

dopeters:
**Summary:** Expand table columns

Addresses [CUMULUS-1885: Expand table columns on double click](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1885)

## Changes

* Column expands to fit content when double clicking the separator/resizer, which is the same behavior that is seen in excel.
* Some style tweaks. Made resizer bigger/easier to click

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - Update Pagination input to show possible page options in dropdown
 
 ### Added
+- **CUMULUS-1885**
+  - Expand columns to fit full content width when separator is double clicked
+
 - **CUMULUS-1895**
   - Update execution events to display more details in modal
 
--- a/app/src/css/modules/_table.scss
+++ b/app/src/css/modules/_table.scss
@@ -20,11 +20,13 @@
     max-width: 500px;
   }
 
+  .thead, .tbody {
+    width: fit-content;
+    min-width: 100%;
+  }
+
   .thead {
     background-color: $ocean-blue;
-    &, .tbody {
-      width: fit-content;
-    }
     .th {
       color: $white;
       font-weight: $base-font-semibold;
--- a/app/src/css/modules/_table.scss
+++ b/app/src/css/modules/_table.scss
@@ -199,7 +199,9 @@
 
 .resizer {
   display: inline-block;
-  background: $lightest-grey;
+  background-color: $lightest-grey;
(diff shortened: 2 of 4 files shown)


jjmccoy:
@dopeters Nice work here. The size tweak to the expander is such a plus! I have one UX question concerning the double-click option. Is it is possible that the target for the double-click be the table header of that column instead of the narrow separator?  I'm thinking more about the ability for someone to have pinpoint accuracy versus more of a target area. If not, possibly adding a hover hint that says "double click here to expand"? Thanks.

dopeters:
> @dopeters Nice work here. The size tweak to the expander is such a plus! I have one UX question concerning the double-click option. Is it is possible that the target for the double-click be the table header of that column instead of the narrow separator? I'm thinking more about the ability for someone to have pinpoint accuracy versus more of a target area. If not, possibly adding a hover hint that says "double click here to expand"? Thanks.

So my thinking here was to match the Excel functionality, which has the double click on the divider. I was also worried about conflicting with the sort click in the header. However, Excel does give a larger target area, and it is still hard to click exactly on our divider so I agree that we should address that.

I pushed a change that adds hover text as mentioned. I also added padding around the divider that gives a larger target area and makes it so you don't have to click exactly on the line. Let me know if that is an acceptable solution. Happy to work it into the main header if not!

jjmccoy:
> > @dopeters Nice work here. The size tweak to the expander is such a plus! I have one UX question concerning the double-click option. Is it is possible that the target for the double-click be the table header of that column instead of the narrow separator? I'm thinking more about the ability for someone to have pinpoint accuracy versus more of a target area. If not, possibly adding a hover hint that says "double click here to expand"? Thanks.
> 
> So my thinking here was to match the Excel functionality, which has the double click on the divider. I was also worried about conflicting with the sort click in the header. However, Excel does give a larger target area, and it is still hard to click exactly on our divider so I agree that we should address that.
> 
👍 Gotcha about the sort click interaction and Excel functionality match. 

> I pushed a change that adds hover text as mentioned. I also added padding around the divider that gives a larger target area and makes it so you don't have to click exactly on the line. Let me know if that is an acceptable solution. Happy to work it into the main header if not!
@dopeters Thanks for this. The padding helped and it is a workable solution.

jjmccoy (approved):
LGTG!

PR #913, opened 2020-12-21
[CUMULUS-2322] Pagination bugfix

dopeters:
**Summary:** Fixed pagination bug

Addresses [CUMULUS-2322: Pagination in rec reports not working and causing page to rerender](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2322)

## Changes

* Removed `<form>` wrapper from `SortableTable` so that button elements in pagination don't trigger a form submit. The form wrapper isn't doing anything so it's safe to remove.
* Also removed `node-notifier` so we can close out https://github.com/nasa/cumulus-dashboard/pull/912. We are no longer using this dependency so we can remove it.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - **CUMULUS-2263**
   - Update Pagination input to show possible page options in dropdown
 
+- **CUMULUS-2322**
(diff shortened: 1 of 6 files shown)


kkelly51 (approved):
Looks great!

PR #919, opened 2021-01-07
Migrate to Cypress 6.2.1

flamingbear:
Addresses [CUMULUS-2325: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2325)

## Changes

* Migration to specs that upgrade Cypress tests to use 6.2.1
* Main changes is to replace `cy.route` & `cy.server` with `cy.intercept`.
* Followed the Migration Guide https://docs.cypress.io/guides/references/migration-guide.html#Migrating-cy-route-to-cy-intercept 

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [X] Integration tests

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -9,6 +9,5 @@
   },
   "nodeVersion": "system",
   "viewportWidth": 1300,
-  "viewportHeight": 1000,
-  "experimentalNetworkStubbing": true
+  "viewportHeight": 1000
 }
--- a/cypress/integration/collections_spec.js
+++ b/cypress/integration/collections_spec.js
@@ -23,12 +23,11 @@ describe('Dashboard Collections Page', () => {
     beforeEach(() => {
       cy.login();
       cy.visit('/');
-      cy.server();
-      cy.route('POST', '/collections').as('postCollection');
-      cy.route('GET', '/collections?limit=*').as('getCollections');
-      cy.route('GET', '/collections/active?limit=*').as('getActiveCollections');
-      cy.route('GET', '/collections?name=*').as('getCollection');
-      cy.route('GET', '/granules?limit=*').as('getGranules');
+      cy.intercept('POST', '/collections').as('postCollection');
+      cy.intercept('GET', '/collections?limit=*').as('getCollections');
+      cy.intercept('GET', '/collections/active?limit=*').as('getActiveCollections');
+      cy.intercept('GET', '/collections?name=*').as('getCollection');
+      cy.intercept('GET', '/granules?limit=*').as('getGranules');
     });
 
     it('should display a link to view collections', () => {
@@ -106,11 +105,10 @@ describe('Dashboard Collections Page', () => {
     });
 
     it('should display expected MMT Links for a collections list', () => {
-      cy.route({
-        method: 'GET',
-        url: '/collections?limit=*',
-        response: 'fixture:collections-with-mmtLinks.json',
-      }).as('getCollections');
+      cy.intercept(
+        { method: 'GET', url: new RegExp('/collections\\?limit=.*') },
+        { fixture: 'collections-with-mmtLinks.json' }
+      ).as('getCollections');
       cy.visit('/collections');
       cy.clearStartDateTime();
       cy.wait('@getCollections');
@@ -154,10 +152,10 @@ describe('Dashboard Collections Page', () => {
 
       // After POSTing the new collection, make sure we GET it back
       cy.wait('@postCollection')
-        .then((xhr) => cy.request({
+        .then((interception) => cy.request({
           method: 'GET',
-          url: `${new URL(xhr.url).origin}/collections/${name}/${version}`,
-          headers: xhr.request.headers
+          url: `${new URL(interception.request.url).origin}/collections/${name}/${version}`,
+          headers: { Authorization: interception.request.headers.authorization },
         }))
         .then((response) => {
           cy.expectDeepEqualButNewer(response.body, collection);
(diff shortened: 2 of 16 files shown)


cypress.json
```diff
@@ -9,6 +9,5 @@
   },
   "nodeVersion": "system",
   "viewportWidth": 1300,
-  "viewportHeight": 1000,
-  "experimentalNetworkStubbing": true
```
flamingbear, line 13: This is the default now

cypress/integration/collections_spec.js
```diff
@@ -106,11 +105,10 @@ describe('Dashboard Collections Page', () => {
     });
 
     it('should display expected MMT Links for a collections list', () => {
-      cy.route({
-        method: 'GET',
-        url: '/collections?limit=*',
-        response: 'fixture:collections-with-mmtLinks.json',
-      }).as('getCollections');
+      cy.intercept(
+        { method: 'GET', url: new RegExp('/collections\\?limit=.*') },
```
flamingbear, line 109: `cy.intercept` doesn't seem to use string globbing like cy.route did.  So any of the urls with asterisks need to be regexps.

cypress/integration/granules_spec.js
```diff
@@ -38,88 +38,93 @@ describe('Dashboard Granules Page', () => {
       // shows a list of granules
       cy.getFakeApiFixture('granules').as('granulesListFixture');
 
+      const numGranulesToTest = 4;
```
flamingbear, line 41: This was just an optimization, this test is really slow, and repeated for every granule in the fixture, this just tests the first 4 instead of all 10.  but takes about 90secs off of the test time.

cypress/integration/granules_spec.js
```diff
@@ -634,33 +628,20 @@ describe('Dashboard Granules Page', () => {
       cy.contains('button', 'Delete').click();
       cy.contains('.button--submit', 'Confirm').click();
 
-      cy.route2({
```
flamingbear, line 637: This was never being hit before, and I verified that changing the property test to a different value that the tests still passed.  The new cy.intercept on L609-614, intercepts the request tests the body to have correct action, and then passes back a success now.

cypress/integration/main_page_spec.js
```diff
@@ -213,7 +213,8 @@ describe('Dashboard Home Page', () => {
       // Cypress only allows one stub per url. We make multiple POST requests to the same
       // elasticsearch endpoint. The fixture here returns a combined response of all the
       // responses for one url, effectively stubbing our elasticsearch searches.
-      cy.route('POST', 'http://example.com/_search/', 'fixture:elasticsearch.json');
+      let fixtureName = 'elasticsearch.json';
+      cy.intercept('POST', 'http://example.com/_search/', (req) => req.reply({ statusCode: 200, fixture: fixtureName }));
```
flamingbear, line 217: `cy.intercept` doesn't allow you to overwrite an intercepted URL (https://stackoverflow.com/questions/65044704/change-fixture-response-in-cypress-for-the-same-url-with-intercept). So this is a way to change the fixture.  It's seems inelegant, but I don't have a better solution.

cypress/integration/pdrs_spec.js
```diff
@@ -223,12 +221,11 @@ describe('Dashboard PDRs Page', () => {
             .should('have.text', `${Number(granule.duration.toFixed(2))}s`);
           cy.get('@columns').eq(5).invoke('text')
             .should('match', /.+ago$/);
-
-          cy.route('DELETE', `/granules/${granule.granuleId}`).as('deleteGranule');
+          cy.intercept('DELETE', `/granules/${granule.granuleId}`).as(`deleteGranule${granule.granuleId}`);
```
flamingbear, line 224: Also, like above, you can't reassign an alias and then wait for it, so here we use unique aliases.

dopeters (approved):
This is awesome! Glad we can get rid of all of those `cy.server()` calls. Lots of good updates to make some of these work and loving the reduced time for the granules test. Just had a question for my own clarification, but all tests are passing, so approved!

cypress/integration/collections_spec.js
flamingbear: That's a great question.  I think there is something different happening in the two cases.  It's not documented, but I see different behavior in testing.

So the documentation says that a string type url input, uses minimatch as the matcher.

`url = "http://localhost:5001/collections?limit=50&page=1&sort_key%5B%5D=-duration&getMMT=true&includeStats=true"
`
`Cypress.minimatch(url, '/collections?limit=*') => false
`
and the `minimatch.makeRe(url) => /^(?:\/(?=.)collections[^/]limit=[^/]*?)$/`

Which doesn't match the url
`url.match(/^(?:\/(?=.)collections[^/]limit=[^/]*?)$/) => null
`

But on extra testing, I could have switched the url to `'**/collections?limit=*'`:
`Cypress.minimatch(url, '**/collections?limit=*') => true
`

So to make sure, I changed L28 to this:
```
cy.intercept({method:'GET', url:'/collections/active?limit=*'}).as('getActiveCollections');
```

And that fails the first test, when it waits for that alias.

TL:TL; 
I think I don't want to change the existing RegExps even though they could be simplified, but I did verify that the intercepts on L26-30 are correctly being intercepted, in the given form, but when a routeMatcher object is provided, those same glob-strings fail.

flamingbear: I might post an issue with Cypress, because I don't think I'm misreading the documentation.

flamingbear: Posted an issue to Cypress, but it turns out it's probably my problem. When you use a routeMatcher with that style of match, you need to set matchUrlAgainstPath: true like this:
```
      cy.intercept({method:'GET', url:'/collections/active?limit=*', matchUrlAgainstPath: true }).as('getActiveCollections');
```
But I will not go back and change them since a RegExp is also acceptable. 
Issue https://github.com/cypress-io/cypress/issues/14458

dopeters: Cool, thanks for investigating. Seems like it's easier to just use the RegEx than to add an additional param anyway. Good to know there's a way to get the routeM matching working though!

PR #920, opened 2021-01-07
[CUMULUS-2176] Update GNF Report Styles

dopeters:
Date range moved to same line as header
Added Total Conflict Comparisons
Download report below border
Table legend moved to Collapse component above table

Addresses [CUMULUS-2176: Reports: GNF View - Update Style & Interaction](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2176)

## Changes

* Added Total Conflict Comparison section w/ description (for both GNF and Inventory)
* Removed report status
* Moved date range to be in line with header
* Made Legend collapsible and added it next to the column filters collapsible button
* Added tooltips for the colored indicators

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/js/components/Form/_form.scss
+++ b/app/src/js/components/Form/_form.scss
@@ -62,62 +62,6 @@ select option{
     background-color: $ocean-blue;
     box-shadow: inset 2px 2px $ocean-blue;
     color: $white;
-}
-}
-
-
-.filters {
-  margin: 1.5em 0 2em;
-  @extend .clearfix;
-  label {
-    font-size: .86em;
-  }
-}
-
-.filter__item {
-  display: inline-block;
-  vertical-align: bottom;
-  position: relative;
-}
-
-.filter__icon{
-  &:after {
-    color: $ocean-blue;
-    position: absolute;
-    font-family: 'FontAwesome';
-    content: '\f0d7';
-    font-weight: 900;
-    width: 14px;
-    height: 14px;
-    background-size: 14px 14px;
-    display: block;
-    right: .8em;
-    top: .5em;
-  }
-}
-
-.filters.filters__wlabels ul{
-  display: flex;
-  align-items: baseline;
-}
-
-.filters.filters__wlabels .run_bulk{
-  margin-left: auto;
-}
-
-.filter-action-wrapper {
-  display: flex;
-  flex-wrap: wrap;
-  align-items: flex-end;
-}
-
-.filter-actions {
-  display: flex;
-  justify-content: space-around;
-  flex-grow: 1;
-
-  .form--controls {
-    flex-grow: 1;
   }
 }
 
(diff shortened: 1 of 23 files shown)


app/src/js/components/Form/_form.scss
```diff
@@ -62,74 +62,6 @@ select option{
...
-  align-items: flex-end;
-}
-
-.filter-actions {
-  display: flex;
-  justify-content: space-around;
-  flex-grow: 1;
-
-  .form--controls {
-    flex-grow: 1;
```
dopeters, line 132: None of these styles are used anywhere

jjmccoy:
@dopeters Looks good! I did notice one thing: When click on 'Show Column Filters' to make a selection it pushes the Table Legend heading and button to the right. 

Screen Shot 2021-01-11 at 11 13 06 AM

dopeters:
@jjmccoy Yeah I had looked into that but was having trouble getting the buttons to stay put. Seems like the `Collapse` component is intentionally designed that way. It does make it so there's room for the legend to expand when both buttons are clicked. Otherwise there would be some overlap, or the legend wouldn't display under its button. Probably could find a workaround, but was having trouble fighting against the styles that come with the component
Screen Shot 2021-01-11 at 11 26 11 AM

One alternative is making the column options display vertically so it doesn't push the button over like so:
Screen Shot 2021-01-11 at 11 25 50 AM

Could also do something to the effect of making the button width bigger so that the area for the column options is wider, thereby allowing us to wrap them. Something like this:
Screen Shot 2021-01-11 at 11 33 30 AM

Let me know your thoughts on those options or if you have any other ideas here!

jjmccoy:
> @jjmccoy Yeah I had looked into that but was having trouble getting the buttons to stay put. Seems like the `Collapse` component is intentionally designed that way. It does make it so there's room for the legend to expand when both buttons are clicked. Otherwise there would be some overlap, or the legend wouldn't display under its button. Probably could find a workaround, but was having trouble fighting against the styles that come with the component
> Screen Shot 2021-01-11 at 11 26 11 AM
> 
> One alternative is making the column options display vertically so it doesn't push the button over like so:
> Screen Shot 2021-01-11 at 11 25 50 AM
> 
> Could also do something to the effect of making the button width bigger so that the area for the column options is wider, thereby allowing us to wrap them. Something like this:
> Screen Shot 2021-01-11 at 11 33 30 AM
> 
@dopeters Oh, I see. I think this workaround would be good. If we make the 'Show/Hide Column Filters' wider then we may need to put the selections into columns or have more padding when they wrap.
> Let me know your thoughts on those options or if you have any other ideas here!

dopeters:
> @dopeters Oh, I see. I think this workaround would be good. If we make the 'Show/Hide Column Filters' wider then we may need to put the selections into columns or have more padding when they wrap.

Thanks for confirming! Just pushed that change

jjmccoy:
@dopeters Nice! Thanks for the updates.

jjmccoy (approved):
LGTG!

PR #921, opened 2021-01-08
[CUMULUS-2148] Fix report deletion bug

bt6453:

Addresses [CUMULUS-2148: reconciliation reports are not removed from reactTable after deletion (every time)](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2148)

## Changes

* Table checks to make sure report was deleted from Elasticsearch and if not it pulls down the reports again

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [X] Adhoc testing
- [ ] Integration tests


app/src/js/utils/table-config/reconciliation-reports.js
```diff
@@ -85,10 +85,24 @@ const handleCsvDownloadClick = (e, reportName, dispatch) => {
...
-    dispatch(listReconciliationReports(query));
+    refreshReports(dispatch, value, query);
   });
 };
 
+// It can take an extra second or two to delete a report from
+// Elasticsearch. This function keeps checking the reports to make sure it was
+// deleted before displaying the delete result.
+const refreshReports = (dispatch, value, query) => {
+  dispatch(listReconciliationReports(query)).then((reports) => {
```
flamingbear, line 96: This should probably have a hard limit to the number of retries and maybe a very small delay. 

My first testing attempt was to delete a badly named record, and this went into an infinite loop. 
Maybe use p-retry with small backoff or something like that? It seems like it's calling a bunch of times (3-4ish) on a successful delete, maybe start with a 250ms or 500ms delay? what do you think?

flamingbear: You can recreate by creating a report with a name with slashes in it (there is a ticket to fix this).  e.g. "01/20/2021 report name"

flamingbear (changes requested):

app/src/js/utils/table-config/reconciliation-reports.js
bt6453: Ah I thought this would fail like normal. Bad assumption on my part. And yeah it takes 2-4 calls to finally get it right but let me try adding a delay to see if it works out better.

flamingbear: I'm also little concerned that it does fail, but we don't seem to do anything about the failure.  But that's out of scope here.

app/src/js/utils/table-config/reconciliation-reports.js
```diff
@@ -85,24 +85,13 @@ const handleCsvDownloadClick = (e, reportName, dispatch) => {
 const handleDeleteClick = (e, value, dispatch, query) => {
   e.preventDefault();
   dispatch(deleteReconciliationReport(value)).then((response) => {
-    refreshReports(dispatch, value, query);
+    setTimeout(
```
flamingbear, line 88: You're probably going to be irritated with me, but I meant that you should have a small delay in your other solution, I like that it retries (would be nice if it retried a finite number of times....).

app/src/js/utils/table-config/reconciliation-reports.js
```diff
@@ -11,7 +11,7 @@ export const tableColumns = ({ dispatch, isGranules, query }) => ([
     Header: 'Name',
     accessor: 'name',
     Cell: ({ cell: { value }, row: { original: { type } } }) => { // eslint-disable-line react/prop-types
-      const link = (location) => ({ pathname: `/reconciliation-reports/report/${encodeURIComponent(value)}`, search: getPersistentQueryParams(location) });
+      const link = (location) => ({ pathname: `/reconciliation-reports/report/${value}`, search: getPersistentQueryParams(location) });
```
flamingbear, line 14: Looks like a bad merge here?  I think the encodedURI was Jenny's fix for unreadable filenames.   And that's what failed your build it looks like.

bt6453: woops! will fix

flamingbear (approved):

flamingbear:
Looks good to go.  Thanks for humoring me. nice work.

PR #923, opened 2021-01-12
[CUMULUS-1763] Test conditional rendering

dopeters:
**Summary:** Add tests for dashboard rendering that differs based on environment variables

Addresses [CUMULUS-1763: Dashboard should test conditional rendering. that is, rendering based on configuration.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1763)

## Changes

* Added an `ava-conditional.config.cjs` file that passes different env variables to the tests
* Added tests for various components that test both the default behavior and the behavior when certain env variables are passed
* `npm run test` now also separately runs ava with the new config file
* Files are called `condition.js` in the folder of the component that is tested. I don't _love_ this naming convention, so I'm open to suggestions. As long as we're able to include/exclude all of the files in some way (probably based on some shared name or part of their name), it should work

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


jennyhliu (approved):
LGTM

PR #926, opened 2021-01-14
Upgrade webpack and apply minimization in production.

flamingbear:
**Summary:** Allows dashboard to be minimized and build with maven.earthdata docker image cache. 

Addresses [CUMULUS-2259: Dashboard deloyment bundle should not include sensitive content.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2259)

## Changes

* Stop code from using `=>` operator to do `bind`ing  [Just don't do it.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
* Updates production build to minify code

Bonus:
* Adds a prefix to localAPI docker-compose image names of ${DOCKER_REPOSITORY}  This is set to an internal repository on the bamboo system and will pull from there in CI. But will will work locally if the variable is unset because it will default to an empty string. 

## PR Checklist

- [X] Update CHANGELOG
- [n/a ] Unit tests
- [X ] Adhoc testing
- [n/a ] Integration tests


dopeters (approved):
Looks good! Confirmed minimization

PR #930, opened 2021-01-20
CUMULUS-2310:Allow slash in reconciliation report name

jennyhliu:

Addresses [CUMULUS-2310: certain inventory filenames make it impossible to retrieve from the api](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2310)

## Changes

* Allow slash in reconciliation report name

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


flamingbear (approved):
I think you should update the fixture just for completeness sake.
The rest of the code is good.  I verified that I could create, download and open in the dashboard, reports with slashes in them.

PR #935, opened 2021-02-08
CUMULUS-2321: Updated reconciliation report to work with the API endpoint changes

jennyhliu:

Addresses [CUMULUS-2321: reconciliationReport endpoint cannot send an "overlarge" json object](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2321)

## Changes

* reconciliationReport API now returns pre-signed URL and data, so update reconciliation report to work with the API endpoint changes
* If the report data is an error message (payload size exceeds max), Inventory report and GNF pages won't be displayed, instead, the page will display error.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/app/src/js/utils/download-file.js
+++ b/app/src/js/utils/download-file.js
@@ -32,14 +32,6 @@ const convertToCSV = (data, columns) => {
   return `${csvHeader}\r\n${csvData}`;
 };
 
-export function handleDownloadJsonClick(e, { data, reportName }) {
-  e.preventDefault();
-  const jsonHref = `data:text/json;charset=utf-8,${encodeURIComponent(
-    JSON.stringify(data)
-  )}`;
-  downloadFile(jsonHref, `${reportName}.json`);
-}
-
 export function handleDownloadUrlClick(e, { url }) {
   e.preventDefault();
   if (url && window && !window.Cypress) window.open(url);
--- a/app/src/js/utils/download-file.js
+++ b/app/src/js/utils/download-file.js
@@ -40,6 +40,11 @@ export function handleDownloadJsonClick(e, { data, reportName }) {
   downloadFile(jsonHref, `${reportName}.json`);
 }
 
+export function handleDownloadUrlClick(e, { url }) {
+  e.preventDefault();
+  if (url && window && !window.Cypress) window.open(url);
+}
+
 export function handleDownloadCsvClick(e, { reportName, table }) {
   e.preventDefault();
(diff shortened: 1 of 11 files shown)


app/src/js/utils/download-file.js
```diff
@@ -40,6 +40,11 @@ export function handleDownloadJsonClick(e, { data, reportName }) {
   downloadFile(jsonHref, `${reportName}.json`);
 }
```
flamingbear, line 41: handleDownloadJsonClick isn't used anymore, so this can probably be removed.

app/src/js/utils/table-config/reconciliation-reports.js
```diff
@@ -14,9 +14,8 @@ export const tableColumns = ({ dispatch, isGranules, query }) => ([
       const link = (location) => ({ pathname: `/reconciliation-reports/report/${encodeURIComponent(value)}`, search: getPersistentQueryParams(location) });
       switch (type) {
         case 'Internal':
-          return <Link to={link} onClick={(e) => handleDownloadClick(e, value, dispatch)}>{value}</Link>;
         case 'Granule Inventory':
-          return <Link to={link} onClick={(e) => handleCsvDownloadClick(e, value, dispatch)}>{value}</Link>;
+          return <Link to={link} onClick={(e) => handleDownloadClick(e, value, dispatch)}>{value}</Link>;
```
flamingbear, line 18: Since you replaced `handleCsvDownloadClick` with the `handleDownloadClick`, can that be replaced everywhere and the function removed also?

flamingbear (changes requested):
Looks good, I tested with all of the report types, including one that is too large, and validated the error shown, but that I could download the file from the dashboard.  Nice.

One piece of code to remove, and one question about removing other code for simplification, but I'm guessing the answer is no on that second one, I just don't know why.  

I will approve when the dead code is removed. Thanks!

app/src/js/utils/table-config/reconciliation-reports.js
jennyhliu: handleCsvDownloadClick has been removed.  handleDownloadCsvClick is a different function

flamingbear: Oh, that's tricky naming, I didn't even see that. Ok, thanks for looking. I'll approve this.

flamingbear (approved):

flamingbear:
Nice Job

PR #936, opened 2021-02-17
[CUMULUS-2366] Logs Errors

dopeters:
**Summary:** Handles logs endpoint errors. Should be tested with master branch of cumulus (which contains changes from this PR https://github.com/nasa/cumulus/pull/2071)

Addresses [CUMULUS-2366: [Dashboard] Handle logs endpoint errors if Metrics not set](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2366)

## Changes

* Dispatch `getLogs` in the main component to check if metrics are configured. If a "Metrics not configured" error is returned, it updates the state with `metricsNotConfigured: true`. Did it this way rather than use a dashboard env variable to keep the core side as the one source of truth. I.e the core config is what tells us it's not configured not any settings on the dashboard side.
* Updated components to leverage the state to determine whether or not to display logs, including in places where logs are linked, like the sidebar.
* Added tests for all places where logs could show up. Used fixtures here so we can test cases of metrics configured or not.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


app/src/js/components/Executions/execution-status.js
```diff
@@ -1,290 +1,109 @@
...
-        {/* stateMachine's definition and executionHistory's event statuses are needed to draw the graph */}
-        {stateMachine && executionHistory && (
-          <section className="page__section">
-            <div className="heading__wrapper--border">
-              <h2 className="heading--medium with-description">Visual</h2>
-            </div>
-            <ExecutionStatusGraph executionStatus={executionStatus} />
-          </section>
-        )}
-
```
dopeters, line 260: Primary change here is moving metaAccessors into the table-config file and simplifying the modal toggling.

jennyhliu (approved):
Looks great 👍
Noticed that with metrics configuration, the log entries are empty in the dashboard except for the execution details page.  It seems to be existing behavior, I just wrote a ticket for it.
https://bugs.earthdata.nasa.gov/browse/CUMULUS-2419

PR #937, opened 2021-02-26
[CUMULUS-2415] Fix issue with displaying old execution graphs

dopeters:

Addresses [CUMULUS-2415: Executions Graph Not Being Rerendered](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2415)

## Changes

* Instead of only rendering graph on mount, put that logic in a `useEffect` block
* Make `ExecutionStatusGraph` a functional component
* Show loading instead of old execution while waiting to fetch the new one

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


app/src/js/components/Executions/execution-status-graph.js
```diff
@@ -16,93 +16,72 @@ import { window } from '../../utils/browser';
...
+  const [showModal, setShowModal] = useState(false);
 
+  useEffect(() => {
     const workflow = JSON.parse(stateMachine.definition);
     const events = getExecutionEvents(executionHistory);
     const graph = workflowToGraph(workflow);
     addEventsToGraph(events, graph);
-    this.g = draw(graph);
-    this.renderGraph('.visual', this.g);
-  }
```
dopeters, line 44: The biggest issue was that the graph rendering was being done in a `componentDidMount` block. That means that whatever `executionStatus`'s value was when it was first passed is what the final graph will display, even if it changes due a new fetch call.

By making this a functional component and moving that logic into a `useEffect`, it will create a graph using the latest version of `executionStatus`

app/src/js/components/Executions/execution-status.js
```diff
@@ -21,14 +22,20 @@ const ExecutionStatus = ({
...
     dispatch(getCumulusInstanceMetadata());
-  }, [dispatch, match.params]);
+  }, [dispatch, executionArnParam]);
+
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, []);
 
   if (!execution) return null;
+  if (executionArn !== executionArnParam) return <Loading />;
```
dopeters, line 38: Just throws in a loading screen so we're never displaying an old execution

jennyhliu (approved):
Nice job👍  No longer see the old execution graph.

PR #939, opened 2021-03-01
Fix version compatibility test (#938)

dopeters:
The property name (isCompatible) was incorrect.
From https://github.com/nasa/cumulus-dashboard/pull/938


flamingbear (approved):
👍

PR #949, opened 2021-04-07
Cumulus 2430

jjmccoy:

Addresses [CUMULUS-2430: UI - Large Stats Display Card with Tooltip](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2430)

## Changes

* Change KPI status cards on Overview pages:
  - Change styling to new design
  - Add tooltip to display full totals

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


app/src/js/utils/format.js
```diff
@@ -71,6 +71,35 @@ export const tally = (numberstring) => {
...
+  if (
+    (!numberstring && numberstring !== 0) ||
+    numberstring === nullValue ||
+    Number.isNaN(number)
+  ) {
+    return nullValue;
+  }
+
+  if (number >= 0) {
+    return `${numeral(number).format('0,0')}`;
```
dopeters, line 85: I think we can just return the output of the format function like this, since that returns a string
```suggestion
    return numeral(number).format('0,0');
```

dopeters (changes requested):
Looks really great! Might be worth adding a quick test to make sure the hover works. Also, there are some lint errors

app/src/js/utils/format.js
jjmccoy: @dopeters Oh okay. Gotcha.

Regarding tests, I have never written a test so I'm not sure that works. 

Still trying to get my Bamboo access to see the lint errors.

dopeters: Ok no problem. Re: lint errors, you can do `npm run lint` locally and it should them to you

jjmccoy: Awesome, thanks @dopeters!

dopeters (approved):
Looks like everything's passing! Thanks for the change.

PR #951, opened 2021-04-08
CUMULUS-2449 Encode collection versions

dopeters:
**Summary:** Add URI encode/decode to collection version

Addresses [CUMULUS-2449:Cumulus dashboard does not display collections and granules with collection version has a /](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2449)

## Changes

* Primary change is encoding the version name when linking to a collection
* Also added formatting functions for the collections hrefs so that all the encoding is happening in one place.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


jennyhliu (changes requested):
I tested against cumulus-std, and I added two collections with '/' in version, and a dummy granule for one of the collections.  There are issues:
1. The collection can't be deleted
2. Collections subpage doesn't show granules when it should.  Other collections subpages may not work either (such as logs)
Screen Shot 2021-04-14 at 3 09 04 PM

app/src/js/components/Collections/overview.js
```diff
@@ -113,220 +108,184 @@ class CollectionOverview extends React.Component {
...
+                })}
+              >
+                Copy
+              </Link>
+            </li>
+            <li>
+              <Link
+                className="button button--edit button--small button--green"
+                to={(location) => ({
+                  pathname: `/collections/edit/${collectionName}/${collectionVersion}`,
```
jennyhliu, line 191: This pathname is not encoded, why?

dopeters: In this case, `collectionVersion` is taken from the params, so it is already encoded

dopeters:
Thanks @jennyhliu! I captured the additional changes like in the sidebar and for deleting a collection. I also added a subsection of integration tests for the encoding, including editing, adding, and deleting a collection with a forward slash in the version

jennyhliu (commented):
Why is package-lock.json updated when there is no package.json change? You can download baseline package-lock.json, and then npm ci to update local packages.

dopeters:
Ah thanks, reverted the package-lock changes

jennyhliu (approved):
Works great!  👍 Thanks

PR #952, opened 2021-04-08
Misconfigured ES Credentials now show an error on the dashboard.

flamingbear:
**Summary:**
Display an error to the dashboard operator when posts to the Metrics' ElasticSearch (ESROOT) fail instead of failing silently and not displaying metrics.

Addresses [CUMULUS-2282: Incorrect metrics credentials can prevent a user from logging into the dashboard.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2282)

## Changes

* Adds component that displays errors returned from a configured metrics
    endpoint.  Specifically, letting a user know that they are unauthorized
    when trying to retrieve data from ESROOT rather than just not showing the
    data.
* Changes requestMiddleware and handleError.  Existing behavior is retained,
    but when an error has no message in it's response, we look to the error
    message which has useful information to show the user.

## PR Checklist

- [ x ] Update CHANGELOG
- [ x ] Unit tests
- [ x ] Adhoc testing
- [ ] Integration tests

Output from cumulus-std correctly configured:
Screen Shot 2021-04-08 at 5 22 49 PM
Output from cumulus-std when user/password is incorrect:
Screen Shot 2021-04-08 at 5 22 44 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - Changes requestMiddleware and handleError.  Existing behavior is retained,
     but when an error has no message in its response, we look to the error
     message which has useful information to show the user.
-  - Upgrades Cypress testing software to 7.0.0
+  - Upgrades Cypress testing software to 7.0.1
 
 ## [v5.0.0] - 2021-03-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - **CUMULUS-2282**
+  - Adds component that displays errors returned from a configured metrics
+    endpoint.  Specifically, letting a user know that they are unauthorized
+    when trying to retrieve data from ESROOT rather than just not showing the
+    data.
+  - Changes requestMiddleware and handleError.  Existing behavior is retained,
+    but when an error has no message in it's response, we look to the error
+    message which has useful information to show the user.
   - Upgrades Cypress testing software to 7.0.0
 
 ## [v5.0.0] - 2021-03-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     when trying to retrieve data from ESROOT rather than just not showing the
     data.
   - Changes requestMiddleware and handleError.  Existing behavior is retained,
-    but when an error has no message in it's response, we look to the error
+    but when an error has no message in its response, we look to the error
     message which has useful information to show the user.
   - Upgrades Cypress testing software to 7.0.0
 
--- a/app/src/js/components/home.js
+++ b/app/src/js/components/home.js
@@ -22,6 +22,7 @@ import {
   tally,
   seconds
 } from '../utils/format';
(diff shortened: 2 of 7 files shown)


app/src/js/middleware/request.js
```diff
@@ -84,22 +99,21 @@ export const requestMiddleware = ({ dispatch, getState }) => (next) => (action)
...
-              error: data,
+              error,
               requestAction,
               statusCode: status
             },
             next
           );
         }
-        handleError({ id, type, error, requestAction }, next);
+        return handleError({ id, type, error, requestAction }, next);
```
flamingbear, line 114: I'm guessing this was never used? I think that's an error that happens in axios. idk but this is probably what was intended.

test/middleware/request.js
```diff
@@ -191,6 +223,26 @@ test.serial('should dispatch login error action for 401 response', async (t) =>
   t.true(loginErrorStub.called);
 });
 
+test.serial('should NOT dispatch login error action for 401 response when made to a non-APIROOT path', async (t) => {
+  nock('https://example-notapiroot.com')
```
flamingbear, line 227: example.com is the default value for apiRoot in our config, lucky for me.

app/src/js/components/home.js
```diff
@@ -204,6 +221,7 @@ class Home extends React.Component {
           </section>
 
           {this.renderButtonListSection(overview, 'Updates')}
+          {this.renderDistrubutionErrors(dist)}
           {this.renderButtonListSection(distErrorStats, 'Distribution Errors', 'distributionErrors')}
           {this.renderButtonListSection(distSuccessStats, 'Distribution Successes', 'distributionSuccesses')}
```
dopeters, line 226: Just thinking out loud here, but I wonder if it makes sense to somehow encapsulate rendering the error vs the distribution button sections into a single render function. Like, if there's errors, display that, and if not, display the two subsections?

Thinking that might make the html markup more logical. Maybe something to consider if we refactor this to a functional component.

dopeters (changes requested):
Looks great! Just a tiny changelog typo and some ramblings

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - **CUMULUS-2282**
+  - Adds component that displays errors returned from a configured metrics
+    endpoint.  Specifically, letting a user know that they are unauthorized
+    when trying to retrieve data from ESROOT rather than just not showing the
+    data.
+  - Changes requestMiddleware and handleError.  Existing behavior is retained,
+    but when an error has no message in it's response, we look to the error
```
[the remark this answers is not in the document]
flamingbear: Ugh the most embarrassing of typos/mistakes. thanks.

app/src/js/components/home.js
flamingbear: I thought about that, It's possible for both errors and stats to be shown (although I had to think hard at how that would be). And I opted for only displaying the distribution errors.  But it does seem like we could make a component out of the whole thing.

dopeters: Ah that makes it even more interesting. Well, something to consider later perhaps

dopeters (approved):
👍🏻

app/src/js/components/home.js
flamingbear: I do kind feel like we have some technical debt to get to.  I wonder if that will ever happen. I will peek at this again, but merge this now.  Thanks.

PR #953, opened 2021-04-12
refactor overview-num card and remove hover effect

flamingbear:
**Summary:** regenerated the package-lock.json file.  We should discuss if this is what we want, or if we should stick with V1 and npm 6.


flamingbear:
closing since we should stick with npm 6 for the immediate future until we decide to move to 7

PR #955, opened 2021-04-14
Home page refactor + new header.

flamingbear:
Refactors home.js Adds new section header for Distribution Overview. 

Screen Shot 2021-04-15 at 10 53 43 AM

Screen Shot 2021-04-15 at 10 48 15 AM


app/src/js/components/home.js
```diff
@@ -147,6 +147,34 @@ class Home extends React.Component {
     }
   }
 
+  pageSection (identifier, title, children) {
+    return (
+      <section className={`page__section ${identifier}`}>
```
flamingbear, line 152: identifier is probably a bad name since this turns into a classname.

app/src/js/components/home.js
```diff
@@ -147,6 +147,34 @@ class Home extends React.Component {
...
+  sectionHeader (identifier, title, link) {
+    return (
+      <section className={`page__section ${identifier}`}>
+        <div className='row'>
+          <div className='heading__wrapper--border'>
+            <h2 className='heading--large heading--shared-content--right'>{title}</h2>
+            {link}
+          </div>
+        </div>
+      </section>
```
flamingbear, line 174: This is basically the same as a section, but with a --border and large heading. ripe for fixing.

app/src/js/components/home.js
```diff
@@ -199,63 +220,41 @@ class Home extends React.Component {
...
+          {this.buttonListSection(overview, 'Updates')}
 
-          <section className='page__section update--granules'>
-            <div className='row'>
-              <div className='heading__wrapper--border'>
-                <h2 className='heading--large heading--shared-content--right'>Granules Updates</h2>
-                <Link className='link--secondary link--learn-more' to={{ pathname: '/granules', search: searchString }}>{strings.view_granules_overview}</Link>
-              </div>
-            </div>
+          {this.sectionHeader('distribution--overview', 'Distribution Overview')}
```
flamingbear, line 234: This probably shouldn't show up if there's no metrics configured.  I need to fix that.

app/src/js/components/home.js
```diff
@@ -199,63 +220,41 @@ class Home extends React.Component {
...
-              <>{strings.granules_updated}<span className='num-title'>{numGranules}</span></>
-            )}
+          {this.sectionHeader(
+            'update--granules', 'Granules Updates',
+            <Link className='link--secondary link--learn-more' to={{ pathname: '/granules', search: searchString }}>{strings.view_granules_overview}</Link>
+          )}
+          {this.buttonListSection(
+            updated,
+            <>{strings.granules_updated}<span className='num-title'>{numGranules}</span></>
+          )}
```
dopeters, line 246: This is causing an issue in the markup since that jsx header is being passed as the className

`<section class="page__section [object Object]">`

app/src/js/components/home.js
```diff
@@ -147,6 +142,32 @@ class Home extends React.Component {
...
+            <h2 className={`heading${sizeMod} heading--shared-content--right`}>
+              {title}
+            </h2>
+            {link}
+          </div>
+          {children}
+        </div>
+      </section>
+    );
+  }
```
dopeters, line 169: I feel like it might be nice to pull this out of this file and make it a component that can be used across the app

app/src/js/components/home.js
```diff
@@ -147,6 +142,32 @@ class Home extends React.Component {
...
+
+  section ({ className, title, link, border = false, size = 'medium', children }) {
+    const borderMod = border ? '--border' : '';
+    const sizeMod = `--${size}`;
+    return (
+      <section className={`page__section ${className}`}>
+        <div className='row'>
+          <div className={`heading__wrapper${borderMod}`}>
+            <h2 className={`heading${sizeMod} heading--shared-content--right`}>
```
dopeters, line 160: Do we need the `heading--shared-content--right` class? Genuinely don't know, but was playing around with it in the browser and not sure that's it's doing much

dopeters (changes requested):
Overall looks great. Loving how much cleaner this makes the home page. Just had a few notes.

The bad className and Distribution Overview showing when there's no metrics should be fixed.

app/src/js/components/home.js
flamingbear: I can pull those out, I'm not positive where we can replace them elsewhere in the app.  I see things that look similar, but that can be a next step.

app/src/js/components/home.js
flamingbear: I can't really see anyplace it's used except the granule overview page and I can replace that with a sectionHeader now.

app/src/js/components/home.js
flamingbear: 2ab4f8bb

app/src/js/components/home.js
flamingbear: Nope can't actually because the `row` div is causing problems, I don't want to rush though a bad change, so I'm going to leave just the homepage changes and leave some more refactoring for another time.

PR #968, opened 2021-06-10
Cumulus 2554 - Accessibility - Fix Focus for Form Inputs, Links, and Buttons

jjmccoy:

Addresses [CUMULUS-2554: Fix Focus for Keyboard Tab Control Accessibility](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2554)

## Changes
* Applied `:focus-visible` pseudo-element for keyboard and mouse differentiation 
* Added fallbacks for older and non-supported browsers
* The keyboard user can `tab` and `shift-tab` through the page and focus shows for links, buttons, and inputs based on the browsers' version

Background: After research and working on this ticket, I decided to apply the new progressive `:focus-visible` for flexibility of keyboard and mouse users and then added fallbacks for older and non-supported browsers. For browsers that don't support this element, `:focus` and `:-moz-focusring` (Firefox) are being applied - note that keyboard and mouse users will see the same styles and behavior because `:focus-visible` is not supported for them. The resources that I used for this ticket are the following:
- https://www.w3.org/TR/selectors-4/#the-focus-visible-pseudo 
- https://hacks.mozilla.org/2019/06/indicating-focus-to-improve-accessibility/ 
- https://stackoverflow.com/questions/31402576/enable-focus-only-on-keyboard-use-or-tab-press 
- https://css-tricks.com/almanac/selectors/f/focus-visible/ 
- https://matthiasott.com/notes/focus-visible-is-here 

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


dopeters (approved):
Looks great! Huge win for accessibility

PR #972, opened 2021-06-22
CUMULUS-2535

akshayshiv:

Addresses [CUMULUS-2535: Update KPI cards on the executions page](https://bugs.earthdata.nasa.gov/browse/CUMULUS-XXX)

## Changes

* Changed PDR tab to maintain consistency across all tabs
* Imported Overview into the timer class so that the constant can be used every time when the KPI cards need to be updated
* Created a new function so that it can refresh the page and update the cards at the same time

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


dopeters:
@akshayshiv Great work here! Though, I don't think we're quite targeting the root of the issue. I think the main cause of this bug has to do with the fact that we're not calling the `getCount` action when the timer is updated. In `Overview/overview.js` we have this code:

```
  useEffect(() => {
    dispatch(getCount({
      type,
      field: 'status',
      ...params,
      ...queryParams
    }));
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [dispatch, JSON.stringify(params), JSON.stringify(queryParams), type]);
```

Essentially what this does, is when any of the variables in the dependency array ([dispatch, JSON.stringify(params), JSON.stringify(queryParams), type]) change, then `getCount` is called. However, none of those specified variables will change when the timer is updated. I think what we need to do is leverage the `inflight` prop in order to dispatch `getCount` when the timer is updated.

You'll notice that for example in `Executions/overview.js` we're always setting inflight to false.

`<Overview type='executions' inflight={false} />`. 

Instead, we should be able to get the inflight value from `executions.list.inflight` and pass that into the overview component. Then, we can add inflight to the dependency array in `Overview/overview.js` and add an if statement wrapper around the dispatch. (i.e. `if (!inflight) dispatch(getCount) ....`).

That way, once the executions list is updated, it triggers the count update.

Hope that makes sense! Let me know if you have any questions.

app/src/js/components/Executions/overview.js
```diff
@@ -66,7 +66,7 @@ class ExecutionOverview extends React.Component {
               Execution Overview
             </h1>
             {lastUpdated(queriedAt)}
-            <Overview type='executions' inflight={false} />
+            <Overview type='executions' inflight={executions.list.inflight} />
```
dopeters, line 69: Awesome! Can we do something similar in the other places we're using the Overview component?

akshayshiv: Yup, will get those changes out!

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     logs, likewise use a pattern like the ES\_DISTRIBUTION\_TARGET\_PATTERN.
 
 ### Changed
+
+- **CUMULUS-2535**
+  - Ensured that the KPI cards would be updated on the executions tab whenever the page was updated
```
dopeters, line 33: Nit: can we change the wording here to indicate that this update is for all KPI cards?

app/src/js/components/Timer/timer.js
```diff
@@ -105,7 +104,7 @@ class Timer extends React.Component {
         <span
           data-cy="refreshTimer"
           className="form__element__refresh"
-          onClick={() => this.refreshTimer(this.props.config)}
+          onClick={() => this.refreshTimer(this.props.config) }
```
dopeters, line 107: Other nit:
Can we just revert the changes in this file?

dopeters (approved):
Awesome work! Thanks for addressing the feedback. I tested this with my stack and the KPIs were updated as an execution went from running to completed. Just a couple of nits but nothing that should hold up approval

PR #976, opened 2021-06-25
CUMULUS-2534

akshayshiv:
Added a copy button to errors pages so error can be copied and searched up on the internet

Addresses [CUMULUS-2534: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2534)

## Changes

* Added a copy button to errors pages so error can be copied and searched up on the internet

* ...

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [x] Integration tests

### DIFF
--- a/app/src/css/_buttons.scss
+++ b/app/src/css/_buttons.scss
@@ -486,16 +486,12 @@ $delete-btn-hover-bg-color: darken($error-red, 10%);
     &.button--copy {
       &:hover{
         background-color: $midnight-blue;
-    }
+      }
       float: right;
       background-color: $ocean-blue;
       padding: .25em 1.2em .25em 2.5em;
       font-weight: $base-font-regular;
-
-
-
-      
-  }
+    }
 
 
     &__icon--animation {
--- a/app/src/css/_buttons.scss
+++ b/app/src/css/_buttons.scss
@@ -494,6 +494,7 @@ $delete-btn-hover-bg-color: darken($error-red, 10%);
 
 
 
+      
   }
 
 
--- a/app/src/css/_buttons.scss
+++ b/app/src/css/_buttons.scss
@@ -483,6 +483,19 @@ $delete-btn-hover-bg-color: darken($error-red, 10%);
         }
       }
     }
+    &.button--copy {
+      &:hover{
+        background-color: $midnight-blue;
+    }
+      float: right;
+      background-color: $ocean-blue;
+      padding: .25em 1.2em .25em 2.5em;
+      font-weight: $base-font-regular;
+
+
+
+  }
+
 
     &__icon--animation {
       cursor: pointer;
--- a/app/src/css/_buttons.scss
+++ b/app/src/css/_buttons.scss
@@ -487,10 +487,11 @@ $delete-btn-hover-bg-color: darken($error-red, 10%);
       &:hover{
         background-color: $midnight-blue;
       }
-      float: right; 
       background-color: $ocean-blue;
       padding: .25em 1.2em .25em 2.5em;
       font-weight: $base-font-regular;
+      position: relative;
+      left: 30%;
     }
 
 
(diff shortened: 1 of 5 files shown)


app/src/css/_buttons.scss
```diff
@@ -483,6 +483,20 @@ $delete-btn-hover-bg-color: darken($error-red, 10%);
...
     }
+    &.button--copy {
+      &:hover{
+        background-color: $midnight-blue;
+    }
+      float: right;
+      background-color: $ocean-blue;
+      padding: .25em 1.2em .25em 2.5em;
+      font-weight: $base-font-regular;
+
```
dopeters, line 494: Can we get rid of the empty spaces here, please?

app/src/js/components/Errors/copy-to-clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+}) => {
+  const [copy, copiedTextState] = useState('Copy');
+
+  async function handleClick(e) {
+    e.preventDefault();
+    try {
+      await navigator.clipboard.writeText(text);
+      copiedTextState('Copied!');
+    } catch (err) {
+      copiedTextState('Not Copied!');
```
dopeters, line 15: Nit: can we make this more descriptive with something like 'Error copying'. I don't know that we need to display the error itself, but I think it would make it more clear that something happened while trying to copy.

app/src/js/components/Errors/copy-to-clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+          className="button button--small button--primary button--show button--copy"
+          onClick={handleClick}
+        >
+          {copy}
+        </button>
+      </>
+    );
+  }
+
+  return render();
```
dopeters, line 32: Instead of making a render function, can we just return the JSX directly here.

app/src/js/components/Errors/report.js
```diff
@@ -37,6 +38,7 @@ class ErrorReport extends React.Component {
         <div key={report}>
           <strong>Error:</strong>
           <ShowMoreOrLess text={report} truncate={this.props.truncate}></ShowMoreOrLess>
+          <CopyToClipboard text = {report}></CopyToClipboard>
```
dopeters, line 41: Another nit, but can we fix the spacing and make this a self closing tag?

```suggestion
<CopyToClipboard text={report} />
```

(And if you want to do the same thing for the `ShowMoreOrLess` component, I wouldn't complain :) )

dopeters (changes requested):
@akshayshiv Seriously awesome job here!!! Had a few stylistic notes, but the functionality works great!

A couple additional things I noticed:

BatchAsyncCommand is not using the error report component and therefore looks like this, with no copy button:
Screen Shot 2021-06-25 at 2 33 56 PM
Can we try using the ErrorReport component there instead so we get the goodness of the copy functionality? (Here's the line for your reference https://github.com/nasa/cumulus-dashboard/blob/develop/app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js#L258). If it turns out there's some technical issue with doing so, then no worries, but might as well try it out.

Also, just re: the styling of the button itself, I'm not sure if we want it to be inside the error box or if it looks ok where it is. It feels like it might be better to have to contained within the box.
Screen Shot 2021-06-25 at 2 36 27 PM
It would be good to have input from @jjmccoy on that one.

app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
```diff
@@ -254,8 +255,9 @@ export class BatchCommand extends React.Component {
             }
             {status === 'error' &&
             <>
-              <Alert variant="danger"><strong>Error</strong></Alert>
```
dopeters, line 257: We should keep the Alert

dopeters (changes requested):

dopeters (approved):
Thanks for addressing all the feedback! Looks great! We should keep on eye on the different places the button appears in case there some styling issues down the line.

PR #980, opened 2021-07-07
[CUMULUS-2549] Collections table sorting

dopeters:
**Summary:** Make stats and MMT link columns sortable on collections overview page

Addresses [CUMULUS-2549: Dashboard: As a Cumulus Operator, I need to be able to sort every column in the Collections Overview table](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2549)

## Changes

* Created sortable table utils file to contain helper function
* Handle custom sorting methods that are passed to the table config - modified some code from react-table. This can't be overridden when manual sorting is set to true, which is necessary for the server side sorting of the other columns. Instead implemented custom versions for our code

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - Ensured that the KPI cards would be updated on all tabs whenever the page was updated
 - **CUMULUS-2544**
   - Changed the timestamp so that the date and time is shown rather than time elapsed since last occurrence
+- **CUMULUS-2549**
+  - All columns on collections overview page are now sortable
 - **CUMULUS-2551**
   - Added a sortable column to individual granules tab so you can sort each file within the granule by size. 
 - **CUMULUS-2554**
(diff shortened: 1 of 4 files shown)


jjmccoy:
UI/UX Review: Tested all columns in the Collections table and functionality looks good. 
Overall Review: Really like the simplicity of the sortable table util and custom sorting solutions for this complex issue.

jjmccoy:
@dopeters Update: Reviewed a larger set of records (86) with Matt S. since the local only has 6 records and noticed that the columns that are newly able to sort are not sorting the whole set but just sorting what is on the page. MMT column was what flagged us when it wasn't sorting properly for ascending order: It should have been on the last page but was the last record on page one instead.

dopeters:
> @dopeters Update: Reviewed a larger set of records (86) with Matt S. since the local only has 6 records and noticed that the columns that are newly able to sort are not sorting the whole set but just sorting what is on the page. MMT column was what flagged us when it wasn't sorting properly for ascending order: It should have been on the last page but was the last record on page one instead.

Ah yeah thanks for catching this this. Will fix

dopeters:
Closing for now as the ticket is blocked and it is currently not planned work

PR #983, opened 2021-07-20
CUMULUS-2511

akshayshiv:

Addresses [CUMULUS-2511Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2511)

## Changes

* Added a reducer and an action to hold sort state
* Had to edit unit tests to ensure that the new sort state maintenance could carry through
* 
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


app/src/js/actions/index.js
```diff
@@ -944,3 +944,4 @@ export const filterReconciliationReport = (param) => ({ type: types.FILTER_RECON
 export const clearReconciliationReportFilter = (paramKey) => ({ type: types.CLEAR_RECONCILIATION_FILTER, paramKey });
 
 export const toggleSidebar = () => ({ type: types.TOGGLE_SIDEBAR });
+export const sortPersist = () => ({ type: types.SORTS });
```
dopeters, line 947:
```suggestion
export const sortPersist = (sortBy) => ({ type: types.SORTS, sortBy });
```

We want to pass information to the action so that the reducer can do something with it

app/src/js/components/SortableTable/SortableTable.js
```diff
@@ -10,9 +10,11 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import omit from 'lodash/omit';
 import { useTable, useResizeColumns, useFlexLayout, useSortBy, useRowSelect, usePagination } from 'react-table';
+import { dispatch } from 'd3';
```
dopeters, line 13: We don't want to be importing dispatch from d3 because it's not the right method.

I think what might be easiest here is doing

`const dispatch = useDispatch()` in the component and importing useDispatch from `react-redux`

app/src/js/components/SortableTable/SortableTable.js
```diff
@@ -10,9 +10,11 @@ import PropTypes from 'prop-types';
...
 import SimplePagination from '../Pagination/simple-pagination';
 import TableFilters from '../Table/TableFilters';
 import ListFilters from '../ListActions/ListFilters';
+import sortPersist from '../../reducers/sort-persist';
```
dopeters, line 17: We want to import the action not the reducer

app/src/js/components/SortableTable/SortableTable.js
```diff
@@ -173,6 +175,7 @@ const SortableTable = ({
   useEffect(() => {
     if (typeof changeSortProps === 'function') {
       changeSortProps(sortBy);
+      dispatch(sortPersist(true, action));
```
dopeters, line 178: Once we have the action imported, this should look something like this

```suggestion
      dispatch(sortPersist(sortBy));
```

app/src/js/reducers/sort-persist.js
```diff
@@ -0,0 +1,15 @@
...
+} from '../actions/types';
+
+export const initialState = {
+  sorted: false,
+  maintained: false,
+};
+export default createReducer(initialState, {
+  [SORTS]: (state, action) => {
+    state.sorted = true;
+    state.maintained = action.holdPreviousState;
```
dopeters, line 13: We want the reducer to do something like this to the state

```suggestion
  [SORTS]: (state, action) => {
    state[action.table]sortBy = action.sortBy
```

From the table, sortBy looks like this:

```
[
	{
		id: [whateverfield is sorted],
		desc: true/false
	}
]
```

So ideally then, our state will look like this:
```
SORTS: [
	[table]: sortByArray
]

That way, on each table, we can preserve the state and then when we query the state, we can do
sort.collections for example on the collections page to get the sortBy value.

File: app/src/js/components/Table/Table.js [hunk 6]
```diff
@@ -48,10 +50,19 @@ const List = ({
...
-
+  let sortBy;
+  if (tableID) {
+    sortBy = sorts[tableID];
+  }
+  let idMatch;
+  if (!sortBy || sortBy.length === 0) {
+    idMatch = initialSortId;
+  } else {
+    idMatch = sortBy[0].id;
```
dopeters, line 61: We don't just want the first sortBy entry, there could be multiple. We should be passing the full sortBy object to `buildSortKey`. Also worth noting that we can't be sure that desc is always true for `sortBy[0].id`, so this wouldn't work in that case

akshayshiv: I don't understand. How can there be multiple values that you can sortBy, when you change the value that you are sorting by, doesn't it just get replaced? Also why does it being descending matter in this case because we aren't checking for that? I am just a little confused

dopeters: `sortBy` is an array generated by `react-table` and stored in the state (and which we're retrieving here). It looks like this `Array<Object<id: columnId, desc: Bool>>`. There can be multiple objects in this array because `react-table` supported multi column sort.

`initialSortId` is different because it is just defining the default column we want to sort by. This is why we pass that id to `[{ id: initialSortId, desc: true }]` so that we're constructing an array in the same format that the table component is expecting. `sortBy` is already in the correct format, so we don't need to do anything to it.

akshayshiv: Got it, pushing changes again

app/src/js/components/BatchAsyncCommands/BatchAsyncCommands.js
```diff
@@ -263,7 +265,11 @@ export class BatchCommand extends React.Component {
             }
             {(!inflight && !status) && (!modalOptions || !modalOptions.children) && (
               <>
-                <div>{confirmText}</div>
+                <div>{confirmTextArray.map((confirmText, index) => (
+                  <React.Fragment key={index}>
+                    {confirmText}
+                  </React.Fragment>
+                ))}</div>
```
dopeters, line 272: Made this change because there was a console warning about a lack of unique keys for elements in a list

kkelly51 (approved):
Checks out

PR #989, opened 2021-08-11
Prevent windows from changing line endings.

flamingbear:
-------

Addresses [CUMULUS-2616: bootcamp update](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2616)

## Changes

* Adds .gitattribute file to prevent windows from changing the line endings.
 https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 * text eol=lf
+*.png binary
+*.pdf binary
+*.ttf binary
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
markdboyd, line 1: You probably want to add some directives to avoid modifying binary files, like

```
# Denote all files that are truly binary and should not be modified.
*.png binary
*.jpg binary
```

npauzenga (approved):
If Menno says it's good... 😄

PR #1000, opened 2021-09-23
downgrade docker cluster to use elasticsearch that matches AWS version 5.3

flamingbear:
**Summary:**
Changes docker container versions of elasticsearch from 5.6 => 5.3 to match AWS.
Addresses CUMULUS-bugfix: CI problems, but also we should just do this.


kkelly51 (approved):
Fine

PR #1024, opened 2021-11-19
[CUMULUS-2749] Fix env variable reading

dopeters:

Addresses [CUMULUS-2749: Fix reading of environment variables on the Dashboard](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2749)

## Changes

* Added `computeBool` function to handle bool env variables that come in as strings

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


app/src/js/config/config.js
```diff
@@ -1,10 +1,23 @@
...
+ * computeBool
+ * @description Takes string env variables and converts them to the intended bool value
+ * @param {string} value value to convert to boolean
+ * @param {boolean || string} defaultValue default value for variable if none provided by user
+ * @returns boolean value or default
+ */
+function computeBool (value, defaultValue) {
+  if (!value) return defaultValue;
+  if (typeof value !== 'string') return value;
+  return value && value.toLowerCase() === 'true';
```
dopeters, line 11: Using the defaultValue here because otherwise if the value passed in is `false`, then it will automatically fall back to the defaultValue. We want to use `false` when it is passed in by the user

cypress/integration/main_page_spec.js
```diff
@@ -247,16 +245,14 @@ describe('Dashboard Home Page', () => {
...
+      cy.get('.overview-num__wrapper-home > ul#distributionErrors > :nth-child(2)').contains('14');
       cy.get('.overview-num__wrapper-home > ul#distributionErrors > :nth-child(1)').contains('16');
 
-      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(5)').contains('11');
-      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(4)').contains('13');
-      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(3)').contains('15');
-      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(2)').contains('19');
+      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(4)').contains('11');
+      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(3)').contains('13');
+      cy.get('.overview-num__wrapper-home > ul#distributionSuccesses > :nth-child(2)').contains('15');
```
dopeters, line 255: On Bamboo, dashboard is configured with SHOW_TEA_METRICS=false
Since this was being read as a string, it was displaying that tile. Now that the variable is read correctly, it will not show up, so I removed it from this test

kkelly51 (approved):
👍

PR #1026, opened 2021-11-23
Package updates

dopeters:
Updates dashboard to Webpack 5!


jaortega527 (approved):
Looks good.

.eslintrc
```diff
@@ -21,7 +21,7 @@
     "lodash",
     "jsx-a11y"
   ],
-  "parser": "babel-eslint",
+  "parser": "@babel/eslint-parser",
```
jaortega527, line 24: Looks good.

package.json
```diff
@@ -194,16 +195,17 @@
...
     "redux-thunk": "^2.2.0",
     "shortid": "^2.2.16",
     "slugify": "^1.4.6",
+    "stream-browserify": "^3.0.0",
     "stubborn-queue": "^2.0.0",
     "url": "^0.11.0"
   },
   "optionalDependencies": {
-    "cypress": "8.6.0"
+    "cypress": "^8.7.0"
```
jaortega527, line 209: Do we want the `^` on cypress? That will allow cypress to update to any 8.x.x version.

webpack.dev.js
```diff
@@ -39,6 +40,7 @@ const DevConfig = merge.smartStrategy(
   },
   plugins: [
     new webpack.HotModuleReplacementPlugin(),
+    new ESLintPlugin()
```
jaortega527, line 43: It may be just my environment, but anytime I run `HIDE_PDR=false SHOW_DISTRIBUTION_API_METRICS=true ENABLE_RECOVERY=true ESROOT=http://example.com APIROOT=http://localhost:5001 npm run serve` using `webpack.dev.js` I receive the error below.

```
Compiled with problems:

ERROR

../../.eslintrc.js#overrides[0] » eslint-config-airbnb-typescript/lib/shared:
	Configuration for rule "@typescript-eslint/camelcase" is invalid:
	Value {"properties":"never","ignoreDestructuring":false,"ignoreImports":false,"ignoreGlobals":false} should NOT have additional properties.
```
If it's just a local issue then I don't want to hold this PR up.

package.json
dopeters: Good call! Removed

webpack.dev.js
dopeters: Have not been able to repro this on my end. Also doesn't seem to be an issue when running on CI. Worth keeping note of though. It's particularly strange that it calls out typescript when we are not using typescript on the dashboard.

jaortega527: That's a fair point. I don't want to hold up the PR because of some janky VSCode setting

jaortega527: Especially when I was able to build, deploy and run all test prior to me deleting `node_modules` and `package-lock.json`

jaortega527 (approved):
Looks great.

PR #1029, opened 2021-12-06
[CUMULUS-2747] Show relevant create provider fields

dopeters:

Addresses [CUMULUS-2747: Create Provider page should only show needed inputs based on protocol](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2747)

## Changes

* Updated create provider page to only show inputs that are relevant for a particular protocol.
	  * SFTP is the only protocol with Private Key and Cm Key Id
	  * HTTP/HTTPS have Allowed Redirects and Certificate Uri
	  * All but S3 have Port, Username, and Password fields

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/app/src/js/components/Providers/add.js
+++ b/app/src/js/components/Providers/add.js
@@ -21,9 +21,9 @@ const AddProvider = ({ providers }) => {
   const handleInputChange = {
     protocol: (value) => {
       if (value === 's3') setExclude(s3Excludes);
-      else if (!value.includes('http') && !value.includes('sftp')) setExclude([...nonHttpsExcludes, ...nonSftpExcludes]);
+      else if (value.includes('http')) setExclude(nonSftpExcludes);
       else if (value === 'sftp') setExclude(nonHttpsExcludes);
-      else setExclude([]);
+      else setExclude([...nonHttpsExcludes, ...nonSftpExcludes]);
       setSelectedProtocol(value);
     },
   };
--- a/app/src/js/components/Providers/add.js
+++ b/app/src/js/components/Providers/add.js
@@ -54,14 +54,11 @@ const AddProvider = ({ providers }) => {
 };
 
 AddProvider.propTypes = {
-  dispatch: PropTypes.func,
-  providers: PropTypes.object,
-  schema: PropTypes.object
+  providers: PropTypes.object
 };
 
 export default withRouter(
   connect((state) => ({
-    providers: state.providers,
-    schema: state.schema,
+    providers: state.providers
   }))(AddProvider)
 );
--- a/app/src/js/components/Providers/add.js
+++ b/app/src/js/components/Providers/add.js
@@ -3,7 +3,6 @@ import { Helmet } from 'react-helmet';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import PropTypes from 'prop-types';
-// import isEqual from 'lodash/isEqual';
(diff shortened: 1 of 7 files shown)


app/src/js/components/Providers/add.js
```diff
@@ -9,23 +9,55 @@ import { isValidProvider } from '../../utils/validate';
...
-);
+const nonSftpExcludes = ['privateKey', 'cmKeyId'];
+const nonHttpsExcludes = ['allowedRedirects', 'certificateUri'];
+const s3Excludes = [
+  ...nonSftpExcludes,
+  ...nonHttpsExcludes,
+  'username',
+  'password',
+  'port',
+];
```
flamingbear, line 20: Can you make a quick ticket to document these and then do that ticket too? 😬

app/src/js/components/Providers/add.js
```diff
@@ -9,23 +9,55 @@ import { isValidProvider } from '../../utils/validate';
...
+const AddProvider = ({ providers }) => {
+  const [exclude, setExclude] = useState(nonSftpExcludes);
+  const [selectedProtocol, setSelectedProtocol] = useState('http');
+
+  const handleInputChange = {
+    protocol: (value) => {
+      if (value === 's3') setExclude(s3Excludes);
+      else if (value.includes('http')) setExclude(nonSftpExcludes);
+      else if (value === 'sftp') setExclude(nonHttpsExcludes);
+      else setExclude([...nonHttpsExcludes, ...nonSftpExcludes]);
```
flamingbear, line 31: Is this else statement the same as`value === 'ftp'`. can you make it explicit with no default?  

This did seem a little awkward figuring out what items to exclude rather than what additional items to include with each different protocol.
I would have just...nevermind, looking at how addRecord works, I don't think having includes would be workable.

flamingbear (changes requested):
Looks good, I had one question about whether the fields should persist when the protocol gets changed and the field is no longer shown.  e.g. if protocol is http and you have a value in CertificateURI, change protocol to ftp and then back.  Should that persist?   I think it would be nice, but might expand the scope beyond what's reasonable here.

Other than that, I do think the documentation for the protocols needs to be updated on the core side.  either as part of this ticket or a new ticket.  docs : https://nasa.github.io/cumulus/docs/operator-docs/provider

app/src/js/components/Providers/add.js
dopeters: After digging through the core docs, I realized the info for providers was moved to the data management types doc. I updated that here: https://github.com/nasa/cumulus/pull/2604
I also opened a quick PR to add some missing fields to the api docs, which is here: https://github.com/nasa/cumulus-api/pull/316

app/src/js/components/Providers/add.js
dopeters: Yeah that's a good call. Made that change. 

And yeah, agreed that the includes would be more intuitive, but then we'd have to include absolutely every field for each one and I was trying to avoid that, especially since fields may get added, etc.

app/src/js/components/Providers/add.js
flamingbear: Thanks for getting those.

flamingbear (approved):

flamingbear:
Looks good, thanks for the updates

PR #1034, opened 2021-12-17
Cumulus-2543: Persist Filter Criteria

jaortega527:
**Summary:** Restore filter criteria when returning to a view with filters

Addresses [CUMULUS-2543: As a Cumulus Operator, I want my search and filtering criteria to follow along with me when I drill down into a child page and then come back to the parent page](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2543)

## Changes

* Added location-query-params reducer to track query params for every route
* Updated Breadcrumbs, Header (menu), and sidebar to dynamically update <Link> to values with stored query params for given routes.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


app/src/js/components/Collections/collection-header.js
```diff
@@ -25,7 +25,7 @@ const CollectionHeader = ({
     },
     {
       label: 'Collections',
-      href: '/collections',
+      href: '/collections/all',
```
jaortega527, line 28: Changed to `/collections/all` because `/collections` redirects to `/collections/all` and all query params were stored for `/collections/all`, but for <Link>s to `/collections` params were not appended.

app/src/js/components/Collections/overview.js
```diff
@@ -56,7 +56,7 @@ const breadcrumbConfig = [
   },
   {
     label: 'Collections',
-    href: '/collections',
+    href: '/collections/all',
```
jaortega527, line 59: Changed to /collections/all because /collections redirects to /collections/all and all query params were stored for /collections/all, but for s to /collections params were not appended.

app/src/js/components/Header/header.js
```diff
@@ -17,7 +17,7 @@ import { kibanaAllLogsLink } from '../../utils/kibana';
 import { getPersistentQueryParams } from '../../utils/url-helper';
 
 const paths = [
-  [strings.collections, '/collections'],
+  [strings.collections, '/collections/all'],
```
jaortega527, line 20: Changed to /collections/all because /collections redirects to /collections/all and all query params were stored for /collections/all, but for s to /collections params were not appended.

cypress/integration/executions_spec.js
```diff
@@ -84,7 +84,7 @@ describe('Dashboard Executions Page', () => {
       );
 
       cy.contains('nav li a', 'Executions').as('executions');
-      cy.get('@executions').should('have.attr', 'href', '/executions');
+      cy.get('@executions').should('have.attr', 'href').and('include', '/executions');
```
jaortega527, line 87: persisted `limit=x&page=y` parameters meant the href no longer only contains `/executions`.

jaortega527:
I added a test for each major area in the even that a routing change occurs. It's possible the <Link> in the breadcrumb, sidebar or menu could be affected and query parameters might not be persisted.

jjmccoy:
@jaortega527 Wow. Great work here with the search and filter parameters!

jjmccoy:
I noticed an intermittent issue with this test. I ran it again it's not erroring out but just FYI.
collections_test_issue

jjmccoy (approved):
LGTG! Nice work.

jaortega527:
> I noticed an intermittent issue with this test. I ran it again it's not erroring out but just FYI. collections_test_issue

I did see this a few times while working on this, but noticed it happened when running and re-running tests frequently because we re-seed the DB between each test and I saw that when reseed was called while a reseed was in progress an error was thrown and the test(s) would fail because the search in the test would not return any records. It's something I wanted to look at next week, but didn't want to hold this ticket up for that.

PR #1135, opened 2024-06-28
Cumulus 3624

vindigo:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/js/components/oauth.js
+++ b/app/src/js/components/oauth.js
@@ -41,6 +41,7 @@ class OAuth extends React.Component {
 
   componentDidMount () {
     const { token } = this.props.queryParams;
+    // tbd
     if (token) {
       const { dispatch } = this.props;
       this.setState({ token }, () => dispatch(login(token)));
--- a/app/src/js/components/oauth.js
+++ b/app/src/js/components/oauth.js
@@ -41,7 +41,8 @@ class OAuth extends React.Component {
 
   componentDidMount () {
     const { token } = this.props.queryParams;
-    // tbd
+
+    console.log('00000000 token ', token);
     if (token) {
       const { dispatch } = this.props;
       this.setState({ token }, () => dispatch(login(token)));
--- a/app/src/js/components/home.js
+++ b/app/src/js/components/home.js
@@ -25,6 +25,7 @@ import linkToKibana from '../utils/kibana';
 import DatepickerRange from './Datepicker/DatepickerRange';
 import { strings } from './locale';
 import { getPersistentQueryParams } from '../utils/url-helper';
+// import InactivityModal from './Modal/InactivityModal';
 
 class Home extends React.Component {
   constructor (props) {
@@ -34,6 +35,7 @@ class Home extends React.Component {
 
   componentDidMount () {
     const { dispatch } = this.props;
+    console.log('4444444444 getCumulusInstanceData ', getCumulusInstanceMetadata);
     dispatch(getCumulusInstanceMetadata())
       .then(() => this.query());
   }
@@ -177,6 +179,7 @@ class Home extends React.Component {
             />
           )}
         </div>
+        {/* <InactivityModal /> */}
       </div>
     );
   }
--- a/app/src/js/middleware/request.js
+++ b/app/src/js/middleware/request.js
@@ -2,10 +2,7 @@
 import qs from 'qs';
 import _config from '../config';
 import { CALL_API } from '../actions/types';
-import {
-  configureRequest,
-  addRequestAuthorization
-} from '../actions/helpers';
+import { configureRequest, addRequestAuthorization } from '../actions/helpers';
 import log from '../utils/log';
 import { isValidApiRequestAction } from './validate';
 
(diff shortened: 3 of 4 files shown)


app/src/js/components/Modal/InactivityModal.js
```diff
@@ -0,0 +1,109 @@
...
+
+    return () => {
+      window.removeEventListener('keydown', handleKeypress);
+    };
+  }, [hasModal]);
+
+  // Effect to handle showing the modal after 30 minutes of inactivity
+  useEffect(() => {
+    const checkInactivity = setInterval(() => {
+      if (Date.now() - lastKeyPress > 1800000 && !hasModal) { // 1800000 ms = 30 minutes
```
jennyhliu, line 56: Can we make the inactivity time configurable?

app/src/js/App.js
```diff
@@ -50,6 +50,7 @@ const MainRoutes = () => (
 const App = () => {
   const [store] = useState(ourConfigureStore({}));
   const isLoggedIn = () => store.getState().api.authenticated;
+  console.log('app authenticated ', store);
```
jennyhliu, line 53: Delete the logging to console.

app/src/js/components/oauth.js
```diff
@@ -24,6 +24,7 @@ const OAuth = ({
   queryParams,
 }) => {
   const [token, setToken] = useState(null);
+  console.log('oauth authenticated: ', api);
```
jennyhliu, line 27: Delete the logging

app/src/js/reducers/index.js
```diff
@@ -57,7 +57,7 @@ export const reducers = {
   reconciliationReports,
   recoveryStatus,
   sorts,
-  locationQueryParams,
+  locationQueryParams
```
jennyhliu, line 60: Why do you remove the trailing comma? Trailing commas are preferred in objects.

app/src/js/components/Modal/InactivityModal.js
```diff
@@ -0,0 +1,109 @@
...
+      children = {children}
+      />
+    </div>
+  );
+};
+
+const mapStateToProps = (state) => ({
+  hasModal: state.hasModal,
+  lastKeyPress: state.lastKeyPress,
+  // logoutTimer: state.logoutTimer
```
jennyhliu, line 86: Remove the code if not used, same for Line 92.

jennyhliu (changes requested):
1. Can you add a cypress test to verify the functionality?
2. The Modal says 'We will close this session in 5 minutes. ', but after 5 minutes, the Session is not closed or log me out, and I click 'Continue Session', the session is still there.

app/src/js/components/Modal/InactivityModal.js
jennyhliu: What's the difference between this inactivity timeout and session timeout?
The launchpad session can timeout regardless this inactivity, is it correct?  Does the launchpad Session get refresh when we click 'Continue Session'?

jennyhliu:
Additional comments:
1. The session is still expired seconds after I click 'Continue Session'.
2. The modal window appears in the login page.
Screenshot 2024-09-26 at 3 34 58 PM

app/src/js/components/Modal/InactivityModal.js
vindigo: **Can you add a cypress test to verify the functionality?**
I don't think this is necessary as the other modals do not have tests written for this so it shouldn't be a requirement. If it is necessary I would be happy to do it but will need to be shown how as I normally do not deal with Cypress. 

**The Modal says 'We will close this session in 5 minutes. ', but after 5 minutes, the Session is not closed or log me out, and the modal window is still there. I click 'Continue Session', the modal window closes and the session is still there.**
In progress

**Can we make the inactivity time configurable?**
Done

**What's the difference between this inactivity timeout and session timeout? How do they work together?
The launchpad session can timeout regardless this inactivity, is it correct? Does the launchpad Session get refresh when we click 'Continue Session'?**
Currently the session timeout just keeps track if the user is logged in or not. I am not sure what this timeout is set for. The inactivity modal will keep track of the users activity. If there is no activity after 30 min then a pop up will show and a second timer will start. This second timer is set for 5 min. If the user wishes to continue the session then they will need to interact with the modal and select "Continue session" otherwise the user will be logged out. In the future the session timeout and the inactivity timeout will be merged but this will be addressed in another ticket.

app/src/js/App.js
vindigo: Done

app/src/js/components/oauth.js
vindigo: Done

app/src/js/reducers/index.js
vindigo: I put the comma back. The reason why I removed it is because it is not necessary. This is the way I have been trained.

app/src/js/components/Modal/InactivityModal.js
vindigo: Done

PR #1154, opened 2024-10-16
CUMULUS-3892: Update the home page metrics overview section

Nnaga1:

Addresses [CUMULUS-3892: Dashboard - Home Page Metrics Overview Section Update](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3892)

## Changes

* Code: Update timeframe to the past 24 hours
* UI: Add a label to inform user of the timeframe

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/audit-ci.json
+++ b/audit-ci.json
@@ -2,5 +2,5 @@
   "high": true,
   "pass-enoaudit": true,
   "retry-count": 20,
-  "allowlist": ["braces", "d3-color", "webpack-dev-middleware", "ws", "jsonpath-plus", "http-proxy-middleware"]
+  "allowlist": ["braces", "d3-color", "webpack-dev-middleware", "ws", "jsonpath-plus"]
 }


jjmccoy:
@Nnaga1 This looks good. I believe there is a cypress test for the home page: `main_page_spec.js`.

jjmccoy:
So if the user goes to another page and comes back to the home page, the stats from that page temporarily bring those over to the home page and it refreshes. In this video example, it is Rules (for ingest) that has 34 currently. For this, I would suggest creating a ticket for the issue because it is outside the scope of your ticket but may need to be addressed so that users are not confused or think the information is misleading. 

https://github.com/user-attachments/assets/b3648b09-46f7-4301-b3ac-c8c384388d44

jjmccoy:
> I did some additional testing on the issue you mentioned @jjmccoy, it seems like it does this not only for rules/executions for the overview only, if you go from home -> granules, the inverse will happen. It will show all 0's for the statuses before the actual counts. Same with if you go from a different page back to the home page, the granule errors will flash from being a list of what it was before to what it should be (i.e. it shows an old list for one second before changing to what it actually should show). It seems like in some cases/calls this behavior occurs. I think I will create a ticket detailing this, but I really dont think it is worth prioritizing at all probably, since it only lasts for a second and is somewhat persistent across the dashboard. I'm not going to consider that in the scope of this ticket. I think this ticket: #1153, should be reviewed before this is approved in order to clear the CI issues.

Yeah, I figured it was out of scope and a new ticket needs to be created for the issue.

jjmccoy (approved):
@Nnaga1 Thanks for testing it. LGTG!

jjmccoy:
@Nnaga1 Sorry I thought the ticket was done. I didn't know you were adding more UI stuff to it. I need to test the tooltip you just added. We also add Fontawesome icons to the library and bring them in a certain way for rendering purposes. I'll have to see if that was added.

jjmccoy:
> @Nnaga1 Sorry I thought the ticket was done. I didn't know you were adding more UI stuff to it. I need to test the tooltip you just added. We also add Fontawesome icons to the library and bring them in a certain way for rendering purposes. I'll have to see if that was added.

UX question: Looks like there is a label and a tooltip that says the same thing. Are you replacing the label with the tooltip?

jjmccoy:
> @jjmccoy @jennyhliu suggested to add a tooltip saying that the default is the last 24 hours, what I added is how I think it's done. I couldn't figure out how to get it centered without doing all of that. If you have a suggestion/way for it to be better lmk 👀 I had to undo the call to the sectionHeader function and add the code from there here so I could add it to the header, I was struggling getting it centered. It would always go to a new line if I made the tooltip its own div. Pretty much adding a tooltip the the Overview part saying the default is from the last 24 hours.

@Nnaga1 As UI/UX, I recommend just using the tooltip then. You can do it one or two ways and it should work out fine. I would suggest to wrap the tooltip in a <span> instead of a <div> but it can be done without it. Below are examples of how the code would look:

Screenshot 2024-10-21 at 1 34 48 PM
Screenshot 2024-10-21 at 1 34 03 PM

You will see either way the tooltip works and is centered with the <h2>.
Screenshot 2024-10-21 at 1 32 53 PM

jjmccoy:
> @jjmccoy I added the span thing and changed the wording as you suggested, thanks for the comments, the div is for the header not for the tooltip however so I'll leave that alone, lmk if its good for merging now 🙌 👀

Awesome work! Thanks.

jjmccoy (approved):
LGTG! Thanks for the updates.

jjmccoy:
> @jjmccoy I allowlisted `http-proxy-middleware` since its causing the audit to fail, I'll create a ticket to deal with the vulnerability since its out of scope of this ticket. Do you think thats fine?

Yeah, that's fine. I usually go with the philosophy that the PR should stay within the scope of the ticket. This ticket is ready to be merged and another ticket can address the dependency.

PR #1165, opened 2025-02-04
CUMULUS-3862 - Updated documentation to account for short term React upgrade version dependency issues that arise using npm during install.

BWex-NASA:
**Summary:** 
Updated documentation to account for short term React upgrade version dependency issues that arise using npm during install.

Addresses CUMULUS-3862: Dashboard: Upgrade React Router to v6 (https://bugs.earthdata.nasa.gov/browse/CUMULUS-3862)

## Changes

Updated the readme and changelog.

## PR Checklist

- [*] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This version of the dashboard requires Cumulus API >= v19.2.0-alpha.1 (TBD API r
 
 ### Changed
 
+- **CUMULUS-3862**
+  - Updated documentation to account for short term React upgrade version dependency issues that arise using npm.
+
 - **GitHub Issue 1162**
   - Added documentation for beginners starting from scratch.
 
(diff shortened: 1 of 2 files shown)


jjmccoy (approved):
@BWex-NASA LGTG

doug-newman-nasa:
Do we need to preserve `npm ci`? I've heard differing opinions on this.

jjmccoy:
> Do we need to preserve `npm ci`? I've heard differing opinions on this.

In a normal world yes it would just be `npm ci`. Because of the old dependency that doesn't work with the React that we just upgraded to.  We only want the legacy install and its dependencies for connected-react-router and nothing else. Trying to make sure we don't legacy everything.

It would be `npm ci`, they will get an error about the dependency, and then they need to run `npm install connected-react-router@6.9.3 --legacy-peer-deps`.

BWex-NASA:
> Do we need to preserve `npm ci`? I've heard differing opinions on this.

Yes, we would put back the `npm ci ` code and any other helpful comments once the upgrade to React is done as part of CUMULUS-3862. This change is merely temporary in nature, but worth it for anyone new to the Cumulus Dashboard that wants to hit the ground running.

PR #1190, opened 2025-05-05
allow custom env set inital range

etcart:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/app/src/js/reducers/datepicker.js
+++ b/app/src/js/reducers/datepicker.js
@@ -6,11 +6,22 @@ import {
   DATEPICKER_HOUR_FORMAT,
 } from '../actions/types';
 
+const parseInitialDate = () => {
+  // eslint-disable-next-line no-restricted-globals
+  const initialDateRange = isNaN(process.env.INITIAL_DATE_RANGE)
+    ? process.env.INITIAL_DATE_RANGE
+    : Number(process.env.INITIAL_DATE_RANGE);
+  if (findDateRangeByValue(initialDateRange)) {
+    return initialDateRange;
+  }
+  return 'Custom';
+};
+
 // Also becomes default props for Datepicker
 export const initialState = () => ({
   startDateTime: null,
   endDateTime: null,
-  dateRange: findDateRangeByValue('Custom'),
+  dateRange: findDateRangeByValue(parseInitialDate()),
   hourFormat: '12HR'
 });
 


jennyhliu (changes requested):
We could make the configuration value more flexible, not just from the dropdown value, so any number of days 7, 10, 14 would work.
Suggest to add optional INITIAL_DATE_RANGE_IN_DAYS, and set the initial startDateTime accordingly.

```
export const initialState = () => ({
  ...computeDateTimeDelta(initialdaterangeindays),
  endDateTime: null,
  dateRange: findDateRangeByValue('Custom'),
  hourFormat: '12HR'
});
```

app/src/js/reducers/datepicker.js
```diff
@@ -6,11 +6,22 @@ import {
   DATEPICKER_HOUR_FORMAT,
 } from '../actions/types';
 
+const parseInitialDate = () => {
+  // eslint-disable-next-line no-restricted-globals
+  const initialDateRange = isNaN(process.env.INITIAL_DATE_RANGE)
+    ? process.env.INITIAL_DATE_RANGE
```
jennyhliu, line 12: The configuration is retrieved from config, search AUTH_METHOD for example

etcart:
> We could make the configuration value more flexible, not just from the dropdown value, so any number of days 7, 10, 14 would work. Suggest to add optional INITIAL_DATE_RANGE_IN_DAYS, and set the initial startDateTime accordingly.
> 
> ```
> export const initialState = () => ({
>   ...computeDateTimeDelta(initialdaterangeindays),
>   endDateTime: null,
>   dateRange: findDateRangeByValue('Custom'),
>   hourFormat: '12HR'
> });
> ```

it seems weird to initialize on a parameter that can't later be selected from our dropdown

jennyhliu:
> it seems weird to initialize on a parameter that can't later be selected from our dropdown
It's Custom configuration. LP actually wants to display 2 week data. Dropdown is for convenience. We can't add every possible value DAAC wants to use to dropdown.

app/src/js/reducers/datepicker.js
```diff
@@ -1,18 +1,35 @@
...
 } from '../actions/types';
-
 // Also becomes default props for Datepicker
-export const initialState = () => ({
-  startDateTime: null,
-  endDateTime: null,
-  dateRange: findDateRangeByValue('Custom'),
-  hourFormat: '12HR'
-});
+export const initialState = () => {
```
jennyhliu, line 10: I don't understand why we make this complicated, we only need one line change.
-  startDateTime: null,
+  ...computeDateTimeDelta(_config.defaultTimeRangeInDays),

etcart: I want to not go weird when handling strings, if I *just* do that I'll end up with a weird label if something unexpected is entered. and while I'm at it iut seems like I should handle "Recent". I'll try to simplify this some though

etcart: I've tried to simplify it while having correct labels

jennyhliu: How about write a separate method?
`dateRange: findDateRangeByDays(initialDateRange),
`

jennyhliu: Can we rename INITIAL_DATE_RANGE to INITIAL_DATE_RANGE_IN_DAYS to avoid confusion?

bamboo/set-bamboo-env-variables.sh
```diff
@@ -22,6 +22,7 @@ declare -a param_list=(
     "bamboo_SIT_HIDE_PDR"
     "bamboo_SIT_STAGE"
     "bamboo_SIT_DASHBOARD_BUCKET"
+    "bamboo_INITIAL_DATE_RANGE"
```
jennyhliu, line 25: INITIAL_DATE_RANGE_IN_DAYS

localAPI/docker-compose-dashboard.yml
```diff
@@ -14,6 +14,7 @@ services:
     environment:
       - APIROOT=http://localhost:5001
       - ENABLE_RECOVERY=true
+      - INITIAL_DATE_RANGE_IN_DAYS=Recent
```
jennyhliu, line 17: Should be a number

bamboo/set-bamboo-env-variables.sh
etcart: ah, thanks for catching

test/reducers/initialState.js
```diff
@@ -0,0 +1,67 @@
...
+      value: 'Custom',
+      label: 'Custom',
+    },
+    hourFormat: '12HR',
+    endDateTime: null,
+    startDateTime: null,
+  });
+});
+
+test.serial('initialState is based on INITIAL_DATE_RANGE environment variable', (t) => {
```
jennyhliu, line 35: INITIAL_DATE_RANGE-> INITIAL_DATE_RANGE_IN_DAYS

jennyhliu (approved):
Good work 🌟

PR #1206, opened 2025-06-06
CUMULUS-3849

gacondino:

## Changes

* Added network error modal 
* Added unit test

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


etcart:
> > upon re-creating, what I'm finding is that once offline the dismiss button doesn't _really_ dismiss this modal, it simply pops back up again immediately, which don'ts leave me room to continue looking at whatever data I have on my screen
> 
> @etcart this is because it's set up to check your connection status every 3 seconds- once one is found, the modal disappears. From my understanding, this modal was to prevent further work from being done until a connection is established. However, I can change this so that it doesn't have this functionality.

ok, I think you either need to remove the "dismiss" button and have it self-dismiss when the connection re-appears, or have dismissal remove it entirely (until reconnection and re-disconnection

PR #1208, opened 2025-06-12
CUMULUS-4114

gacondino:

Addresses [CUMULUS-4114: Display all Rules in dashboard regardless of initial date-range query](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4114)

## Changes

* Created allRules in index.js to be imported and called in overview.js instead of listRules

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


jennyhliu (changes requested):
The Rules page should not be affected by date picker at all including the count. It has no date picker on that page.
You can look at Providers page, except you can perform search on rules page.

app/src/js/actions/index.js
```diff
@@ -739,6 +739,15 @@ export const listRules = (options) => (dispatch, getState) => {
   });
 };
```
jennyhliu, line 740: Probably we only need to remove timeFilters from listRules.
https://github.com/nasa/cumulus-dashboard/pull/1208/files#diff-b437e4d6db394a61a39ac02fafdf1ffee8356fe7172db3a0b7e21c87815da4b0R731

etcart (approved):
looks good, displays updated rules since X correctly without wiping the display page on a smaller time-frame

jennyhliu (approved):

PR #1237, opened 2025-09-16
CUMULUS-4223 / CUMULUS-4180 - Updating Cypress to v12.17.4

wisdomaj:
Summary: Updating Cypress to v12.17.4

Addresses Changes
[CUMULUS-4223](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4180) 
[CUMULUS-4180](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4180)

* Migrated cypress specs to new folder/file/config format introduced in cypress 10.0.0
* Moved appstore to the top window so app store persists between tests and avoids intermittent issues where appstore cannot be found
* Upgraded cypress package to 12.17.4
* Added more selective selectors for certain tests since in v12.0.0 the .within() command now rejects more than one subject
* Changed breadcrumb url in granules.cy.js to look for searchShort instead of search variable
* Replaced curl with wget in wait-for-dashboard.sh since curl is no longer included in the cypress docker image

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


cypress/e2e/bulk_granules.cy.js
```diff
@@ -15,7 +15,7 @@ describe('Dashboard Bulk Granules', () => {
       cy.contains('button', 'Granule Actions').click();
       cy.contains('button', 'Run Bulk Granules').click();
 
-      cy.get('.bulk_granules')
```
acyu, line 18: Is there a specific reason for this change? I am not aware of any bulk_granules UI changes, so I am a bit confused by this change.

cypress/e2e/main_page.cy.js
```diff
@@ -115,13 +122,16 @@ describe('Dashboard Home Page', () => {
...
+      cy.get('.datetime__wrapper').within(() => {
         cy.get('h3').should('have.text', 'Date and Time Range');
+      });
+      cy.get('.datetime__range').within(() => {
         cy.setDatepickerDropdown('1 hour');
-
-        cy.url().should('include', 'startDateTime=201503171500');
-        cy.url().should('include', 'endDateTime=201503171600');
       });
+      cy.url().should('include', 'startDateTime=201503171500');
```
acyu, line 132: Is there any possibility of race condition happening here?  Maybe not?  I am just wondering if the assertion can happen before the 1 hour date picker action happens

cypress/e2e/main_page.cy.js
```diff
@@ -217,25 +227,15 @@ describe('Dashboard Home Page', () => {
     });
 
     it('Logging out successfully redirects to the login screen', () => {
-      // Logging to debug intermittent timeouts
-      cy.task('log', 'Start test');
-
```
acyu, line 222: Is there specific reason or request to remove the logging?

localAPI/wait-for-dashboard.sh
```diff
@@ -6,17 +6,17 @@ set -e
 
 cmd="$@"
 
-until curl --connect-timeout 5 -sS http://localhost:5001/version 2> /dev/null | grep 'api_version' > /dev/null; do
+until wget --connect-timeout=5 -q -O - http://localhost:5001/version | grep 'api_version' > /dev/null; do
```
acyu, line 9: It seems this wget command does exactly what the previous curl command did. Is there a reason to switch?

cypress/e2e/bulk_granules.cy.js
wisdomaj: The reason for this change is there's a breaking change in cypress 12.0.0 where the `within()` command now requires a single subject and throws an error if given more than one subject. The `cy.get('.modal-dialog')` selector was returning more than one subject so I had to change it to something more selective.

cypress/e2e/main_page.cy.js
wisdomaj: I just removed it because I didn't think it was needed. I thought it was just left over from somebody troubleshooting previously. I didn't see any intermittent timeouts here. We can add this back if you'd like though.

acyu: I think it's simple enough change that we can add them back in the future if we need to.

PR #1255, opened 2025-12-16
Cbanh/cumulus 4189

cqbanh:

Addresses [CUMULUS-4189: Assess implementation of `Enter` & dashboard button "Search" submitting user queries](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4189)

## Changes

- Added new search button serves as an explicit trigger for database queries, which prevents accidental searches from partial input, allows for input validation before querying. 
- The search implementation no longer calls the API on each keystroke, preventing the performance bottleneck that caused timeout issues. 
- Autocomplete now uses
locally cached Redux data (the current page results) with negligible performance impact. 

- Tested it using the LP snapshot instance in SIT.

## PR Checklist

- [x ] Update CHANGELOG
- [x] Adhoc testing
- [x] Integration tests

### DIFF
--- a/app/src/js/components/Search/search.js
+++ b/app/src/js/components/Search/search.js
@@ -131,26 +131,11 @@ const Search = ({
     return { valid: true };
   }, [minSearchLength]);
 
-  const handleSearchClick = useCallback(() => {
-    if (searchRef.current) {
-      searchRef.current.hideMenu();
-    }
-
-    // Block handleChange from triggering
-    blockHandleChange.current = true;
-    setTimeout(() => {
-      blockHandleChange.current = false;
-    }, 100);
-
-    // Get the actual value from the input element
-    const inputElement = searchRef.current?.inputNode;
-    const currentInputValue = inputElement ? inputElement.value : inputValue;
-    const trimmedValue = currentInputValue.trim();
-
+  const executeSearch = useCallback((value) => {
+    const trimmedValue = (value || '').trim();
     setInputValue(trimmedValue);
 
     if (trimmedValue) {
-      // Validate search term
       const validation = validateSearchTerm(trimmedValue);
       if (!validation.valid) {
         setValidationError(validation.error);
@@ -167,8 +152,24 @@ const Search = ({
       dispatch(clear(paramKey));
       setQueryParams({ ...queryParams, [paramKey]: undefined, page: 1 });
     }
-  }, [searchRef, inputValue, validateSearchTerm, infixBoolean, archived,
-    paramKey, dispatch, action, setQueryParams, queryParams, clear]);
+  }, [action, archived, clear, dispatch, infixBoolean, paramKey, queryParams, setQueryParams, validateSearchTerm]);
+
+  const handleSearchClick = useCallback(() => {
+    if (searchRef.current) {
+      searchRef.current.hideMenu();
+    }
+
+    // Block handleChange from triggering
+    blockHandleChange.current = true;
+    setTimeout(() => {
+      blockHandleChange.current = false;
+    }, 100);
+
+    // Get the actual value from the input element
+    const inputElement = searchRef.current?.getInput();
+    const currentInputValue = inputElement ? inputElement.value : inputValue;
+    executeSearch(currentInputValue);
+  }, [searchRef, inputValue, executeSearch]);
 
   function handleChange(selections) {
     // Don't process onChange if we just clicked search button - prevents double trigger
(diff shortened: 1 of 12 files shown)


app/src/js/components/Search/search.js
```diff
@@ -51,51 +51,161 @@ const Search = ({
...
+      const validation = validateSearchTerm(trimmedValue);
+      if (!validation.valid) {
+        setValidationError(validation.error);
+        return;
+      }
+
+      setValidationError('');
+
+      // Dispatch action AND update URL, reset to page 1
+      dispatch(action(trimmedValue, infixBoolean, archived));
```
wisdomaj: HandleSearch, handleSearchClick, and handleKeyDown have this identical code block:
        
```
if (trimmedQuery) {
        setInputValue(trimmedQuery);

        const validation = validateSearchTerm(trimmedQuery);
        if (!validation.valid) {
          setValidationError(validation.error);
          return;
        }

        setValidationError('');
        dispatch(action(trimmedQuery, infixBoolean, archived));
        setQueryParams({ ...queryParams, [paramKey]: trimmedQuery, page: 1 });
      } else {
        setInputValue('');
        setValidationError('');
        dispatch(clear(paramKey));
        setQueryParams({ ...queryParams, [paramKey]: undefined, page: 1 });
      }
}
```

Could we consolidate these code blocks into a single function so it's not duplicated?

app/src/js/components/Search/search.js
```diff
@@ -51,51 +51,161 @@ const Search = ({
...
+
+        const validation = validateSearchTerm(trimmedQuery);
+        if (!validation.valid) {
+          setValidationError(validation.error);
+          return;
+        }
+
+        setValidationError('');
+        dispatch(action(trimmedQuery, infixBoolean, archived));
+        setQueryParams({ ...queryParams, [paramKey]: trimmedQuery, page: 1 });
```
wisdomaj, line 180: I think we may be triggering two search requests for every search. dispatch is called here to initialize a search request and then setQueryParams is called to update the url. This then triggers this useEffect hook where dispatch is called again:
```
  // Re-execute search when other filters change (status, collection, etc)
  useEffect(() => {
    const currentUrlSearchValue = getInitialValueFromLocation({
      location,
      paramKey,
      queryParams,
    });

    if (currentUrlSearchValue &&
            currentUrlSearchValue === inputValue &&
            inputValue.length >= minSearchLength) {
      dispatch(action(inputValue, infixBoolean, archived));
    }
  }, [queryParams, inputValue, minSearchLength, dispatch, action, infixBoolean, archived, location, paramKey]);
```

wisdomaj: Hey Curtis, I have another PR here to fix the duplicate queries https://github.com/nasa/cumulus-dashboard/pull/1268. Duplicate queries were being triggered because dispatch was called when the search button or enter button was pressed, and then called again when the url updated with new query params. This PR just uses the url query parameter from the url as the source of truth and calls dispatch once. It is going through CICD now. Let me know what you think about these changes.

app/src/js/components/Search/search.js
```diff
@@ -14,19 +13,19 @@ import {
...
+ * @param {boolean} archived Whether to include archived records in search
  */
 
 const Search = ({
   action,
   clear,
   dispatch,
   infixBoolean,
-  archived,
+  archived = false,
```
Nnaga1, line 28: whats the logic behind having this as false initially? just if it isnt set to have it as false from the get-go? or

app/src/js/reducers/granules.js
```diff
@@ -152,7 +152,7 @@ export default createReducer(initialState, {
   [SEARCH_GRANULES]: (state, action) => {
     state.list.params.prefix = action.infixBoolean ? null : action.xfix;
     state.list.params.infix = action.infixBoolean ? action.xfix : null;
-    state.list.params.archived = action.archived;
+    state.list.params.archived = action.archived === false ? 'false' : null;
```
Nnaga1, line 155: why not true instead of null?

Nnaga1 (commented):
This looks really good!, im still in the middle of reviewing but a couple of quick comments:

1.) Is there a way you can make the search bar itself smaller so you can put the search button to the right of it? I know enter works as well but for UI reasons I guess it would look better

2.) can you add some simple units or a cypress suite thing for the new changes you made? I think a unit with some assertions would honestly suffice

app/src/js/components/Search/search.js
Nnaga1: ^ agreed 👍

cqbanh:
> This looks really good!, im still in the middle of reviewing but a couple of quick comments:
> 
> 1.) Is there a way you can make the search bar itself smaller so you can put the search button to the right of it? I know enter works as well but for UI reasons I guess it would look better
> 
> 2.) can you add some simple units or a cypress suite thing for the new changes you made? I think a unit with some assertions would honestly suffice

I adjusted the search button to be on the right side of the search text box and added some Cypress tests.

wisdomaj (approved):

PR #1257, opened 2025-12-18
Cumulus 4298 - Exclude collections from date filtering

rswanso5-nasa:

Addresses [CUMULUS-4298: Exclude collections from date filtering](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4298)

## Changes

* Added checkbox to filtering section of Collections view
* Added dateFilter bool to query params
* Updated checkbox component to accept paramKey
* Updated collections tests

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


wisdomaj:
The date filter seems to partially work for me. I'm not getting all of the results I should though.

I'm running the dashboard like so:
```
npm run start-dashboard
npm run seed-database
```
With the date filter set to false, setting a start date of 6/13/2018, and end date of today I get all 6 collection records (http://localhost:3000/collections/all?dateFilter=false&endDateTime=20260102000000&limit=50&page=1&startDateTime=20180613000000)

After enabling the date filter only 2 records are returned when there should still be 6 since all collections were created after 6/14/18 (http://localhost:3000/collections/all?dateFilter=true&endDateTime=20260102000000&limit=50&page=1&startDateTime=20180613000000)

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
```
Nnaga1, line 8: can you add a space in between line 8 and 9? just for appearance I guess, and also right above that the `### Added` like:

```
## [Unreleased]

### Added

- **CUMULUS-4298**
  - Added checkbox to exclude collections from date filtering.
```

app/src/js/actions/index.js
```diff
@@ -117,7 +117,8 @@ export const checkApiVersion = () => (dispatch, getState) => {
 export const listCollections = (options = {}) => {
   const { listAll = false, getMMT = true, includeStats = true, ...queryOptions } = options;
   return (dispatch, getState) => {
-    const timeFilters = listAll ? {} : fetchCurrentTimeFilters(getState().datepicker);
+    const dateFilter = ['true', undefined].includes(queryOptions?.dateFilter);
```
Nnaga1: is there no false option? or would that just be undefined pretty much, mainly curious

app/src/js/components/Checkbox/Checkbox.js
```diff
@@ -12,32 +15,42 @@ const Checkbox = ({
...
-        type="checkbox" checked={checked}
-        onChange={(e) => { onChange(e.target.checked); }} />
-      <span className="checkmark"></span>
-    </label>
-  </div>
-);
+  paramKey,
+  setQueryParams,
+}) => {
+  function handleChange(value) {
```
Nnaga1, line 21: is this change to make it from a class to functional component?, if so fully supported 🙌

Nnaga1:
One thing I noticed while testing this, try this sequence of events to see:

uncheck the box -> put a time/date filter

it doesn't ignore the filter and does the time/date filtering still, _sometimes_ it intermittently ignores it but most of the time doesn't, I observe this behavior with SERVED_BY_CUMULUS_API=true and false and using SIT dashboard, you may need to look into the collections queries reading the DATE_FILTER value to check whether or not to consider it or something like that

Nnaga1 (changes requested):
some small changes mostly formatting, there might be a bug too (I described in the standalone comment I had)

app/src/js/actions/index.js
rswanso5-nasa: Updated the logic to check for `!== 'false'`.

app/src/js/components/Checkbox/Checkbox.js
rswanso5-nasa: The handleChange function was taken from the datepicker component. It is needed to update the state value when the checkbox is updated. I could be wrong, but I believe that the checkbox is already a function component.

rswanso5-nasa:
> One thing I noticed while testing this, try this sequence of events to see:
> 
> uncheck the box -> put a time/date filter
> 
> it doesn't ignore the filter and does the time/date filtering still, _sometimes_ it intermittently ignores it but most of the time doesn't, I observe this behavior with SERVED_BY_CUMULUS_API=true and false and using SIT dashboard, you may need to look into the collections queries reading the DATE_FILTER value to check whether or not to consider it or something like that

This has been fixed. To document my findings on what is causing this issue:
The datepicker on the collections page contained a onChange function which was calling listCollections manually resulting in a second API call. This duplicate call did not include the query params which contained the date filter variable added in this branch. Removing the onChange function fixed the issue while updating the results as expected when the date is changed. This function needs to be removed from the remaining pages: ie. workflows, executions, granules etc.

Nnaga1:
> > One thing I noticed while testing this, try this sequence of events to see:
> > uncheck the box -> put a time/date filter
> > it doesn't ignore the filter and does the time/date filtering still, _sometimes_ it intermittently ignores it but most of the time doesn't, I observe this behavior with SERVED_BY_CUMULUS_API=true and false and using SIT dashboard, you may need to look into the collections queries reading the DATE_FILTER value to check whether or not to consider it or something like that
> 
> This has been fixed. To document my findings on what is causing this issue: The datepicker on the collections page contained a onChange function which was calling listCollections manually resulting in a second API call. This duplicate call did not include the query params which contained the date filter variable added in this branch. Removing the onChange function fixed the issue while updating the results as expected when the date is changed. This function needs to be removed from the remaining pages: ie. workflows, executions, granules etc.

You're right, I just tested and it works perfectly! Thanks for changing and fixing this 🙌

Nnaga1 (approved):
LGTM 🙌 great work! thanks for addressing my concerns and questions, tested on SIT and the checkbox works perfectly 💯

PR #1270, opened 2026-01-08
CUMULUS-4451: Set default sort order for executions to -updatedAt to avoid slow query in Postgres.

ygliuvt:

Addresses [CUMULUS-4451: Set default sort order for executions to -updatedAt to avoid slow query in Postgres](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4451)

## Changes

* Set default sort order for executions to -updatedAt to avoid slow query in Postgres

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - **CUMULUS-4451**
   - Disable sort executions by name to avoid slow query in Postgres
+  - Set default sort order for executions to -updatedAt to avoid slow query in Postgres
   
 - **CUMULUS-4461**
   - Address qs vulnerability by overriding its version to ^6.14.1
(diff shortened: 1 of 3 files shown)


chris-durbin (approved):

jennyhliu (changes requested):
I think it’s better to change the following, because `ORDER BY updatedAt` works best when there is a time-based filter and is not necessarily optimal in other cases.
https://github.com/nasa/cumulus-dashboard/blob/develop/app/src/js/components/home.js#L53

jennyhliu (approved):
Nice work 👍

ygliuvt:
> I think it’s better to change the following, because `ORDER BY updatedAt` works best when there is a time-based filter and is not necessarily optimal in other cases. https://github.com/nasa/cumulus-dashboard/blob/develop/app/src/js/components/home.js#L53

I think for all use cases of the Dashboard, sorting by updatedAt is more meaningful than cumulus id and the generated sql query is faster for Dashboard use cases.

PR #1281, opened 2026-01-28
CUMULUS-4087 Support Automatic Extension of Cumulus Dashboard User Sessions

wisdomaj:
**Summary:**
Implemented automatic session extension for the Cumulus Dashboard. It has a intelligent auto-refresh as long as the user is active. It allows for a 12-hour session cap with independent inactivity timeout.

Addresses [CUMULUS-4087: Automatic Extension of Cumulus Dashboard User Sessions](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4087)
## Changes

* Automatic Token Refresh: Added a SessionTimeoutModal. Tokens auto-refresh when the user is still active and the token is close to expiring (< 5 minutes remaining) with no user interaction
* 12-Hour Session Cap: Sessions capped at 12 hours from initial login, tracked via JWT iat claim
* Modal only shows when user can take action; expired + cap reached = immediate logout
* Lots of tests added to verify token refresh behavior

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


mason-t-yates (approved):

mason-t-yates:
After today's demo all looks good to me 👍

app/src/js/actions/index.js
```diff
@@ -11,34 +11,46 @@ import { configureRequest } from './helpers';
...
+  maxSessionDurationSeconds
 } = _config;
 
 export const refreshAccessToken = (token) => (dispatch) => {
-  const start = new Date();
-  log('REFRESH_TOKEN_INFLIGHT');
+  // Check if session has exceeded 12-hour cap using token's iat claim
+  const sessionStartSeconds = getSessionStart(token);
+
+  if (sessionStartSeconds && (Math.ceil(Date.now() / 1000) - sessionStartSeconds) > maxSessionDurationSeconds) {
```
Nnaga1: NIT: can you put this: `(Math.ceil(Date.now() / 1000) - sessionStartSeconds) > maxSessionDurationSeconds` in `()`, its a little confusing

app/src/js/actions/index.js
```diff
@@ -11,34 +11,46 @@ import { configureRequest } from './helpers';
...
   apiRoot: root,
   defaultPageLimit,
-  minCompatibleApiVersion
+  minCompatibleApiVersion,
+  maxSessionDurationSeconds
 } = _config;
 
 export const refreshAccessToken = (token) => (dispatch) => {
-  const start = new Date();
-  log('REFRESH_TOKEN_INFLIGHT');
```
Nnaga1, line 27: we don't need this since we have this: `dispatch({ type: types.REFRESH_TOKEN_INFLIGHT });` later down, is that why you're removing it? or is it just unneccessary

app/src/js/actions/index.js
```diff
@@ -11,34 +11,46 @@ import { configureRequest } from './helpers';
...
+  // Backend /refresh endpoint handles both Earthdata and Launchpad auth methods
   const requestConfig = configureRequest({
     method: 'POST',
     url: new URL('refresh', root).href,
     data: { token },
   });
+
   return axios(requestConfig)
-    .then(({ body }) => {
-      const duration = new Date() - start;
```
Nnaga1, line 37: why is this not needed anymore: 
```
      const duration = new Date() - start;
      log('REFRESH_TOKEN', `${duration}ms`);
```
? mainly curious, does the api call just not care since it will refresh/extend it anyways?

app/src/js/actions/index.js
```diff
@@ -628,20 +640,40 @@ export const login = (token) => ({
...
+  return dispatch(deleteToken())
+    .then(() => {
+      console.log('[loginError] Token deleted, dispatching LOGIN_ERROR and redirecting to /auth');
+      return dispatch({ type: 'LOGIN_ERROR', error });
+    })
+    .then(() => {
+      console.log('[loginError] Redirecting to auth page');
+      return historyPushWithQueryParams('/auth');
+    });
+};
```
Nnaga1, line 682: great additional console logging 👍 seems pretty useful

app/src/js/actions/index.js
Nnaga1: also you console log for delete and login but not here, is there a reason for that?

app/src/js/actions/index.js
```diff
@@ -628,20 +640,40 @@ export const login = (token) => ({
...
+      return dispatch({ type: types.DELETE_TOKEN });
+    });
 };
 
-export const loginError = (error) => (dispatch) => dispatch(deleteToken())
-  .then(() => dispatch({ type: 'LOGIN_ERROR', error }))
-  .then(() => historyPushWithQueryParams('/auth'));
+export const loginError = (error) => (dispatch) => {
+  console.error('[loginError] Login failed with error:', error);
+  return dispatch(deleteToken())
```
Nnaga1, line 673: so if login errors it will: ` dispatch(deleteToken())` 
this is intended behavior right? login error -> delete token -> login again, it doesnt deviate from what already happens or is this something new?

app/src/js/components/Modal/modal.js
```diff
@@ -16,7 +16,10 @@ const DefaultModal = ({
   hasConfirmButton = true,
   confirmButtonClass = 'button--submit',
   cancelButtonClass = 'button--cancel',
-  size = 'md'
+  size = 'md',
+  animation = true,
```
Nnaga1, line 20: these seem like good additions:

  animation = true,
  backdrop = true,
  closeButton = true

but since you're adding to defaultModal, it might affect the other ones, i see there are no cypress issues but would be good to validate none of the other ones are adversely affected by these new params

app/src/js/components/Modal/modal.js
```diff
@@ -46,8 +49,10 @@ const DefaultModal = ({
       centered
       size={size}
       aria-labelledby={`modal__${className}`}
+      animation={animation}
+      backdrop={backdrop === false ? 'static' : backdrop}
```
Nnaga1: if its false -> static, if true-> true, whats the difference? is it mainly cosmetic/appearance?

app/src/js/components/oauth.js
```diff
@@ -26,20 +26,26 @@ const OAuth = ({
   const [token, setToken] = useState(null);
 
   useEffect(() => {
+    console.log('[OAuth] effect triggered - api.authenticated:', api.authenticated, 'token:', token);
```
Nnaga1: NIT: do we want to console.log the token? might be security or privacy concerns, not sure tho

app/src/js/config/index.js
```diff
@@ -30,7 +30,24 @@ const baseConfig = {
...
+  // token refresh threshold - refresh token when this many seconds or less remain
+  tokenRefreshThresholdSeconds: 5 * 60, // 5 minutes
+
+  // session duration limit
+  maxSessionDurationSeconds: 12 * 60 * 60, // 12 hours
+
+  // session warning threshold - show warning modal when session is within
+  // this many seconds of reaching the max session duration
+  sessionWarningThresholdSeconds: 20 * 60, // 20 minutes
+
```
Nnaga1, line 44: I agree with these values 👍 I remember you showing them in demo-friday as well and there was no issue

app/src/js/middleware/token.js
```diff
@@ -6,40 +6,46 @@ import { loginError, refreshAccessToken } from '../actions';
...
     const tokenExpiration = get(jwtData, 'exp');
     if (!tokenExpiration) {
+      console.error('[refreshTokenMiddleware] No expiration found in token');
       return dispatch(loginError('Invalid token'));
     }
 
-    // tokenExpiration = date seconds since epoch
-    // Math.ceil(Date.now() / 1000) = now in seconds since epoch
-    if ((tokenExpiration - Math.ceil(Date.now() / 1000)) <= refreshInterval) {
+    const currentTime = Math.ceil(Date.now() / 1000);
```
Nnaga1, line 30: this logic changes basically because you dont use refreshInterval anymore from config, you use the other values you added right?

Nnaga1: refreshThresholdSeconds is basically refreshInterval in a sense if im understanding correctly

app/src/js/components/SessionTimeoutModal/session-timeout-modal.js
```diff
@@ -1,70 +1,196 @@
...
 const SessionTimeoutModal = ({
+  token,
   tokenExpiration,
   title = 'Session Expiration Warning',
   children = 'Your session will expire in 5 minutes. Please re-login if you would like to stay signed in.',
   dispatch,
 }) => {
   const [hasModal, setHasModal] = useState(false);
-  const [modalClosed, setModalClosed] = useState(false);
+  // 'sessionCapReached', 'sessionCapWarning', 'tokenExpired', 'tokenExpiringWarning'
```
Nnaga1: NIT: move this comment: `// 'sessionCapReached', 'sessionCapWarning', 'tokenExpired', 'tokenExpiringWarning'` down to the switch/case section

app/src/js/components/SessionTimeoutModal/session-timeout-modal.js
```diff
@@ -1,70 +1,196 @@
...
+        return;
       }
-    }, 1000);
+
+      // If token is expiring soon (but not expired yet)
+      if (tokenSecondsLeft <= _config.sessionWarningThresholdSeconds && tokenSecondsLeft > 0) {
+        // If session cap not reached, auto-refresh
+        // Note: Inactivity is handled separately by InactivityModal
+        if (!sessionCapReached && !refreshAttemptedRef.current && !hasModal && !modalClosedRef.current) {
+          console.log(`[SessionTimeoutModal] Token expiring warning: will expire in ${tokenSecondsLeft}s, session duration: ${sessionDurationSeconds}s`);
```
Nnaga1, line 143: should we console.log here that a refreshAttempt will occurr? Like: 'Token refresh request submitted" or something like that

Nnaga1 (commented):
Some comments I have as review. Great extensive testing + comments which help make it a lot easier to understand. The additional console.logging is also great. I think im good to approve once these NITs and questions are resolved. Great work!! 🙌

app/src/js/actions/index.js
wisdomaj: `log('REFRESH_TOKEN_INFLIGHT');` I think was accidentally deleted. We currently log `deleteToken` and `loginError` but not when the token is refreshed. I will add that logging back. Good catch.

`dispatch({ type: types.REFRESH_TOKEN_INFLIGHT });` was moved to below the below this block: 
```
  // Check if session has exceeded 12-hour cap using token's iat claim
  const sessionStartSeconds = getSessionStart(token);

  if (sessionStartSeconds && ((Math.ceil(Date.now() / 1000) - sessionStartSeconds) > maxSessionDurationSeconds)) {
    const error = new Error('Session has exceeded maximum duration of 12 hours');
    dispatch({
      type: types.REFRESH_TOKEN_ERROR,
      error
    });
    return Promise.reject(error);
  }
```
This is because we only want to actually refresh the token and change the state to inflight if the max session duration hasn't been reached.

app/src/js/actions/index.js
wisdomaj: This logs how long the request took. It's pretty fine grained logging but will add it back. Not sure what my thinking was at the time. Some of the logging, especially in the session timeout modal is pretty verbose and logs often. I added a lot of the logging just to be able to debug things during development and to be able to show for the demo. Before merging this PR it's probably best that I only log when the dashboard is configured for non-production environment by default.

app/src/js/actions/index.js
wisdomaj: Yep! that's intended. I actually did not change any logic here, just the formatting is different since I added some more logging. The original `loginError` function calls `deleteToken` as well, it's just on the same line.

If login fails it's possible the token is invalid, expired, or rejected by the server. We want to delete the token and start with a fresh slate.

If there is a login error the user will be redirected back to the /auth page to login again where we should get a brand new token.

Nnaga1: perfect, thanks for the explanation 🙌

app/src/js/components/Modal/modal.js
wisdomaj: I did make sure `animation` and `backdrop` in DefaultModal were set the same values as the defaults for Modal. The `closeButton` value is being set to true instead of false though. I don't think other modals were designed not to be closed but I will change the default for this to false to align with previous behavior.

app/src/js/components/Modal/modal.js
wisdomaj: `backdrop` controls whether there's a dark overlay behind the modal.

`'static'` makes the background visible but not clickable. This is needed for when the users session expires since we don't want to allow the user to dimiss the modal and want to force the user to re-login.

Now that I think about it it's probably better to set `'static'` in the session timeout modal and have it passed through to the Modal component rather than set the default value this way directly in the Modal component. I will make this change

app/src/js/components/SessionTimeoutModal/session-timeout-modal.js
wisdomaj: We actually already covered that case since the `refreshAccessToken` function logs `Token refresh initiated`. It's called below this comment.

app/src/js/components/oauth.js
wisdomaj: No we don't want to log that, that's a security issue. Good catch! I wll remove the token from the log. I think this was added while debugging.

app/src/js/middleware/token.js
wisdomaj: Yes that's correct, `refreshThresholdSeconds` replaces `refreshInterval`. It just determines how far before token expiration we should refresh the token.

`refreshInterval` was also hardcoded so I made it configurable in the `app/src/js/config/index.js` file instead

Nnaga1 (approved):
Thanks for addressing my feedback! This is some great work as usual! I think this comment may or may not be unresolved (I think you did add it back, if thats the case it can be ignored) : https://github.com/nasa/cumulus-dashboard/pull/1281#discussion_r2819825426 but it was pretty small and don't want to block this PR on that, again, great work 🙌

Nnaga1:
> Thanks for addressing my feedback! This is some great work as usual! I think this comment may or may not be unresolved (I think you did add it back, if thats the case it can be ignored) : [#1281 (comment)](https://github.com/nasa/cumulus-dashboard/pull/1281#discussion_r2819825426) but it was pretty small and don't want to block this PR on that, again, great work 🙌

Also seems like theres some CI issues but it seems irrespective to your PR 👀 (like the audit and pdrs failure things)

app/src/js/actions/index.js
wisdomaj: Logging message for initiating the token refresh has been added back here 

```
console.log('[refreshAccessToken] Token refresh initiated');
dispatch({ type: types.REFRESH_TOKEN_INFLIGHT });
```

